PR #19, opened 2016-03-02
Added a extension method `As<T>` to explicitly cast a object to T

zhenlineo:
Usage example:

```
        foreach (var record in cursor.Stream())
        {
            var node = record["movie"].As<INode>();
            movies.Add(new Movie{
              released = node["released"].As<int>(),
              title = node["title"].As<string>(),
              actors = node["actors"].As<List<string>>() });
        }
```


zhenlineo:
I will merge this one to fix the red build first. The API part we could revisit latter.

PR #26, opened 2016-03-19
Adding Error extension method to the ILogger interface

IngvarKofoed:
Adding Error extension method to the ILogger interface to make it easier to create error log entries when there is no exception instance.


zhenlineo:
Any special reason that this extension method for logging is needed? 
The expected way of using `ILogger` for a user of this driver is to implement `ILogger` so that they could log the error/info/whatever to somewhere they like, e.g. console or a log file. 
If I understand correctly, the one who would call the extension method is only the driver developers, like me and you. No one should call this method outside this driver code.

I have not get the need to have such an extension method, do you think it is needed to have such a method? Do you need it somewhere? Could you give an use case?

IngvarKofoed:
No I can not find any examples right now, so you are right, there is no need for it :)

PR #27, opened 2016-03-20
Fixing culture issues in some of the unittests that failed on non EN enviroments

IngvarKofoed:
Some of the unittest were failing because Windows was not configured with EN locale. This PR fixes this issue.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tck.Tests/TCK/CypherRecordParser.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tck.Tests/TCK/CypherRecordParser.cs
@@ -1,10 +1,10 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text.RegularExpressions;
 using Neo4j.Driver.Extensions;
 using Neo4j.Driver.Internal;
+using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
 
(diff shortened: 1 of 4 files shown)


zhenlineo:
Could you re-base this? Also I was worried if this is a general problem with `Convert.ToDoulbe`, `Convert.ToFloat` method. If so, then maybe we also need to update all other code where this method is called.

IngvarKofoed:
Yes, Ill do that and run through code for other possible issues like this :)

IngvarKofoed:
Rebased done! 

I went through all the code in all projects looking for calls to the methods: `Convert.ToDouble, Convert.ToDecimal, float.Parse, double.Parse, decimal.Parse, float.TryParse, double.TryParse, decimal.TryParse`. There were no more calls that needed invariant culture argument :)

zhenlineo:
Two other files also used the convert for decimals:

Got a PR for you [here](https://github.com/IngvarKofoed/neo4j-dotnet-driver/pull/1)

IngvarKofoed:
I actually do not think that they are needed (which is why I leaf it unchanged) in PackStream.cs because the `Convert` is not converting from string which is the only case where `CultureInfo.InvariantCulture` is used, but for completeness I think it makes sense to have it there :)

Regarding `ValueExtensions`. I though that, like PackStream.cs also only was used internally and only for parsing values from the Neo4j server and assumed that the response was culture invariant. But I see that I was mistaken about the `ValueExtensions` which is for public use. So nice catch :)

PR #29, opened 2016-03-21
Force the client to wait for 10s for the server to fully start

zhenlineo:


jexp:
Shouldn't the driver rather poll the server to see if it is already up? At least that's how I understood @benbc how the new startup-mechanism is supposed to work?

benbc:
Yes! That's much more reliable and quicker.

benbc:
Here's an example of how to do that polling (written in Bash and adapted from the Docker image).

``` bash
end="$((SECONDS+10))"
while true; do
    http_code="$(curl --silent --write-out %{http_code} --output /dev/null http://localhost:7474/db/data/ || true)"

    [[ "${http_code}" = "200" ]]; then
        break;
    fi

    if [[ "${SECONDS}" -ge "${end}" ]]; then
        (cat data/log/console.log && echo "Neo4j failed to start" && exit 1)
    fi

    sleep 1
done
```

zhenlineo:
All right, I tried to do the poll, and then I lost in the asyc socket connection calls in C#. As you guys like to poll to be more reliable, then I will try to do the poll then.

zhenlineo:
Use #33 instead

PR #30, opened 2016-03-22
Making the intergration test fixture os independed

IngvarKofoed:
I have added a Process based Neo4jInstaller that does not use PS or Windows Services to install/uninstall/start/stop the Neo4j services.

This new ProcessBasedNeo4jInstall can be opted in by using the compile symbol PROCESS_TEST_BAESED. If this symbol is not specified, the code run as before.

The motivation for the process based is that it is less error prone due to missing administrator rights and missing PS (Miscrosoft.Automation) assemblies. And it will also work on non-Windows systems, like Linux or OSX using mono.


IngvarKofoed:
Ill update the code to match 1.1 to make the PR easier :)

zhenlineo:
I am a bit surprised that this process installer could run on other os with the server in the windows zip. I actually would suggest instead of using the direct java command to run it, we could download the server in unix tar file and then use the `neo4j start` command normally.

IngvarKofoed:
Yes, that seems like a better solution. I will first update this PR to work with 1.1 and then I can start working on the Mono track and look into the unix tar. I actually started working on getting it to compile in Mono on OSX (including the test projects), but I could not run the intergration tests and didn't want to make any changes without the tests working :)

Please let me know if you think this process installer is a dead end and you have no plans to merge it :)

IngvarKofoed:
Do you have a link for the unix tar file?

zhenlineo:
All the latest downloads could be found here http://alpha.neohq.net/

I would say pls do the changes to get your code working so that you will not be blocked by failed ITs. The ITs are not write for any other platform other than windows at the beginning. And pls issue the Pr against 1.0 as we would like to support multi platforms for the first version too.

But instead of using the pure java command, I would suggest you to do something similar to this

https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/test/java/org/neo4j/driver/v1/util/Neo4jInstaller.java
https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/test/java/org/neo4j/driver/v1/util/Neo4jDefaultInstaller.java (which is a Unix installer, where you got all the commands for running the server)

I got some problem to find a good untar lib so I have not get a PR for your problem yet. 

BTW, would you mind to sign CLA so that we could add your name in our build, the build is red is mainly because the build is not recognizing you now. http://neo4j.com/developer/cla/
Pls let me know when you sign the CLA then I will add your name on the white list. 

I am recently ill and I will take a few days off. I will be back on Next Tuesday. During this time I might be really slow to help you.

GitHub Logo

IngvarKofoed:
I have refactured the code to use the neo4j.bat instead of running the java commend :)

zhenlineo:
I opened a new PR targeting at 1.0. There is no need to also keep a PR for 1.1 as I will do forward merge once the code to previous branches merged in :)
I actually thinking that the process-installer might not be needed. Could you try the new service installer in #33 to see if it works for you? I also change to use `neo4j.bat`, though we still need some code for other os. Once #33 is green, I will merge it in and hopefully you will not be blocked by ITs.

IngvarKofoed:
Yes, Ill try #33 and see if it works without the process stuff :) 

I have already made code and changes so the ITs works on OSX using Xamarin and Mono and it seems to work great. Ill make a PR for this after #33 has been integrated :)

PR #31, opened 2016-03-23
Windows script for running all unittests from a console added for those who do not have ReSharper

IngvarKofoed:
Simple script for running all unittest that also handles the problem with some tests that can't run in parallel. Just makes it easier for running all the tests :)


zhenlineo:
Would mind to re-open the PR to 1.1 instead. Sorry, I just discussed with my collea, and we think this is a nice new feature to have, but not belongs to bug-fix only 1.0 now.

IngvarKofoed:
Sure, no problem! :)

IngvarKofoed:
Done, see #34 :)

PR #32, opened 2016-03-23
Windows script for running all unittests from a console added for those who do not have ReSharper

IngvarKofoed:
Simple script for running all unittest that also handles the problem with some tests that can't run in parallel. Just makes it easier for running all the tests :)


zhenlineo:
No need to targeting 1.1 if you already have a PR to 1.0 :) so close this one

PR #33, opened 2016-03-24
1.0 intergration test fixture os independed

zhenlineo:
Based on https://github.com/neo4j/neo4j-dotnet-driver/pull/30
I just changed two lines 
1) Made the download file location from `{Version}.zip` to `{Version}{platformEnding}`
2) Force to wait in `WindowsServiceBasedNeo4jInstaller.start` too

Then I formated the code to use 4-space-tab. Please use 4-space-tab when contributing, this is the only format requirement at this moment :)

Reopen the PR targeting 1.0 instead


zhenlineo:
@IngvarKofoed I got some problem to use enterprise windows zip file. So I also changed back to community. Would you mind to check out his branch and verify if you could use `WindowsServiceBasedNeo4jInstaller` now? If you could use the new one, do you mind me to remove the process installer?

IngvarKofoed:
I have tested #33 and it works for both community and enterprise :)

I agree that ee do not need the process based installer more now that the other is working. So feel free to delete it :)

Ill make PR for an installer that works on OSX (Linux and Unix) later today or tomorrow

Issue #25, opened 2016-03-19
Unittests made with MSTest which is not supported in Xamarin


IngvarKofoed:
I suggest using NUnit instead of MSTest for the unittest projects. This would make it possible for developers to open and run tests using Xamarin when developing on non Windows platforms.

If you agree, then I can make a pull request where the projects have been changed to use NUnit and tested in both Visual Studio and Xamarin :)

cskardon:
It's using XUnit, not MSTest? Doesn't xunit work with xamarin? A brief search on google shows that it should...

Are you opening the project in xamarin studio?

IngvarKofoed:
The problems is that the projects are created within Visual Studio so they will get an "Test" project type and Xamarin Studio can't handle this and will not load the project. The solution is to change the project type from MSTest project to NUnit project.

Yes, I open the solution in Xamarin Studio: http://imgur.com/VGH1HTO

More information about it here: https://forums.xamarin.com/discussion/22097/xamarin-studio-is-unable-to-open-test-project-created-in-visual-studio-2013

cskardon:
Right, so you don't need to change to nunit, just remove the test guid from the csproj?

IngvarKofoed:
That will fix the issue in XS. Ill try that and test how that works in VS and if it works, then create a PR :)

cskardon:
I'm pretty certain that all it does is not show the test 'icon', so it should work fine!

IngvarKofoed:
Yeah, I was hoping for the same thing, but it was not as easy. I had to create new projects because that Guid was not the only thing that made it not compile in Xamarin on OSX. I have all things compiling and all unittest working. Im working on running integration tests.

Some of the unittests fails because of Windows locale settings. Ill also fix those.

IngvarKofoed:
This can be closed if PR #35 is merged :)

zhenlineo:
Some updates on this issue:

The PR is red due to some lib missing (.NET Framework 4.5.1 SDK) on teamcity windows machine. So working on the machine right now.

zhenlineo:
Hi 

Sorry for not come back for this PR for a long time. 

We looked into this issue and we find the support for compiling source code/run all unittests with Xamarin is not part of our current target. You are most welcome to issue a PR to add this support. But we currently will not add/maintain this feature. 

Cheer,
Zhen

zhenlineo:
[...]

PR #35, opened 2016-03-25
Xamarin and Mono support for integration tests

IngvarKofoed:
Unix based Neo4jInstaller implemented and the used conditional depending on which platform the code in the intergration test is compiled (Mono vs MSbuild/VS). 

I have run all unit tests with success for both community and enterprise editions on OSX after I compiled the soluion in Xamarin. 

Please disragard the Windows script for running all unittest. I have removed it from the branch and this PR.

Enjoy :)

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/IntegrationTestFixture.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/IntegrationTestFixture.cs
@@ -23,7 +23,7 @@ namespace Neo4j.Driver.IntegrationTests
 {
     public class IntegrationTestFixture : IDisposable
     {
-        private readonly INeo4jInstaller _installer = new WindowsNeo4jInstaller();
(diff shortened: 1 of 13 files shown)


zhenlineo:
The test build is really panic about this change, I need more time to make sure it both works on mac and windows and find a good way to verity it in our test build chain for multi-platform support.

I will come back to this PR when I sort out why the test build is not compiling due to the changes in Neo4j.Driver.Tests.csproj and Neo4j.Driver.IntegrationTests.csproj. Also I need a mono test build to verify the support too.

I am also thinking if this should be a 1.1 feature to support compiling source code on multi-platforms. The current goal is to support compiled code working on multi-platforms.

zhenlineo:
I am closing this PR as it is too far behind the current code. As I commented in issue #25, you are most welcome to restart one. The new 1.1 driver is targeting at the new .Net core platform.

PR #36, opened 2016-03-26
Fx for equality in Neo4j entities (INode, IRelationship, etc) that could lead to bugs

IngvarKofoed:
I have changed the code for equality in all entities because the current implementation required that the two instances that one wanted to compare has the same type. Example:

```
class Node : INode {} 
class MyNode : INode {}
var node = new Node(1); // Id = 1
var myNode = new MyNode(1); // Id = 1
var isEqual = node.Equals(myNode); // isEqual would now be false.
```

I have also moved the IEquatable to the entity interfaces because the interfaces are the public part of the driver and it make sense to compare to entities of the same interface.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Entity.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Entity.cs
@@ -73,8 +73,8 @@ namespace Neo4j.Driver.Internal
 
         public bool Equals(IRelationship other)
         {
-            if (other == null)
-                if (ReferenceEquals(this, other)) return true;
+            if (other == null) return false;
+            if (ReferenceEquals(this, other)) return true;
             return Equals(Id, other.Id);
         }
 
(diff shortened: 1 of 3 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Entity.cs
```diff
@@ -74,15 +73,14 @@ public class Relationship : IRelationship, IEquatable<IRelationship>
 
         public bool Equals(IRelationship other)
         {
+            if (other == null)
```
zhenlineo: Did you forget `if(other == null) return false` here?

zhenlineo:
I just lost interment for three days and I got so many PRs now :D
Thanks so much! Beside the comments, this looks good to me.

IngvarKofoed:
No worries! :)
Woops, yes, there should be a null check there. Thanks! :)

zhenlineo:
the null check fix [here](https://github.com/IngvarKofoed/neo4j-dotnet-driver/pull/2)

IngvarKofoed:
Thanks! Merged :)

PR #37, opened 2016-03-26
ILogger is no longer inheriting IDisposable because this should be up the the client of the Neo4j driver to decide

IngvarKofoed:
I have removed the IDisposable from the ILogger interface because I think it is the client of the Neo4j driver that owns its logger and how it wants to use it. So I think it would be wrong if the Neo4j driver disposed it and therefor IDisposable is no longer needed on the ILogger interface.

This is backwards compatible breaking, this is why I created the PR agains 1.0 :)

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Driver.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Driver.cs
@@ -61,7 +61,6 @@ namespace Neo4j.Driver
                 return;
             _sessionPool?.Dispose();
             _sessionPool = null;
-            Logger?.Dispose();
         }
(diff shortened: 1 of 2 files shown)


zhenlineo:
I agree with you. Nice catch!
But you are targeting at 1.1, isn't it? 
IngvarKofoed  wants to merge 1 commit into **neo4j:1.1** from IngvarKofoed:1.0-ilogger-should-not-be-idisposable
Could you re-issue it to 1.0?

IngvarKofoed:
Doh! Yes, sure, done #43

Issue #39, opened 2016-03-27
IStatementResult needs to be IDisposable because of IPeekingEnumerator


IngvarKofoed:
I think you should reconsider if IStatementResult _has_ to be IDisposable because it makes the client code that uses the Neo4j driver more complex and possible needless harder to use. 

I suggest to refactor the IPeekingEnumerator or the use of it so IStatementResult does not have to be IDisposable to make the user code less complex.

If you agree, Ill be happy to make a PR for a suggestion on how to do this :)

zhenlineo:
I would close this issue as we have discussed on https://github.com/neo4j/neo4j-dotnet-driver/pull/50 and the PR now has been merged.
Pls be free to re-open this issue if you feel there is more to discuss.

PR #42, opened 2016-03-27
Fixing compiler warnings

IngvarKofoed:
I have fixed all compiler warnings in all projects because it leave a better impression when cloning and compiling the code :)

Take special note to the commit: https://github.com/neo4j/neo4j-dotnet-driver/commit/e968381c75f1248fe7e00787a7cbca52f1a7b3b2

I think this change is for the better, because the changed classes now behaves like you would expect. Eg: Dispose does not have any side effects other than the object being disposed. And calling most of the methods on a disposed object throws `ObjectDisposedException` because the state of an disposed is undefined.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Result/PeekingEnumeratorTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Result/PeekingEnumeratorTests.cs
@@ -115,15 +115,32 @@ namespace Neo4j.Driver.Tests
         public class DisposeMethod
         {
             [Fact]
-            public void ShouldNotDiscardAfterDispose()
+            public void ShouldDiscardAfterDispose()
             {
                 var list = new List<string> { "1", "2", "3" }.GetEnumerator();
                 var enumerator = new PeekingEnumerator<string>(list);
-                enumerator.MoveNext().Should().BeTrue();
-                enumerator.Consume();
+                enumerator.Dispose();
                 enumerator.MoveNext().Should().BeFalse();
                 enumerator.Peek().Should().BeNull();
-                enumerator.Position.Should().Be(3);
+                enumerator.Position.Should().Be(-1);
+            }
+
+            [Fact]
+            public void PeakShouldReturnDefaultAfterDispose()
+            {
+                var list = new List<string> { "1", "2", "3" }.GetEnumerator();
+                var enumerator = new PeekingEnumerator<string>(list);
+                enumerator.Dispose();
+                enumerator.Peek().Should().Be(null); // null is default(string)
+            }
+
+            [Fact]
+            public void ConsumeShouldThrowAfterDispose()
+            {
+                var list = new List<string> { "1", "2", "3" }.GetEnumerator();
+                var enumerator = new PeekingEnumerator<string>(list);
+                enumerator.Dispose();
+                Assert.Throws<ObjectDisposedException>(() => enumerator.Consume());
             }
         }
     }
--- a/Neo4j.Driver/Neo4j.Driver.Tests/PackStream/PackStreamMessageFormatV1Tests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/PackStream/PackStreamMessageFormatV1Tests.cs
@@ -694,11 +694,11 @@ namespace Neo4j.Driver.Tests
                     {
                         public static INode Alice = new Node(1001L,
                             new List<string> {"Person", "Employee"},
-                            new Dictionary<string, object> {{"name", "Alice"}, {"age", 33l}});
+                            new Dictionary<string, object> {{"name", "Alice"}, {"age", 33L}});
 
                         public static INode Bob = new Node(1002L,
                             new List<string> {"Person", "Employee"},
-                            new Dictionary<string, object> {{"name", "Bob"}, {"age", 44l}});
+                            new Dictionary<string, object> {{"name", "Bob"}, {"age", 44L}});
 
                         public static INode Carol = new Node(
                             1003L,
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Result/StatementResultTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Result/StatementResultTests.cs
@@ -385,26 +385,53 @@ namespace Neo4j.Driver.Tests
         public class DisposeMethod
         {
             [Fact]
-            public void ShouldConsumeRecordStream()
+            public void ShouldNotConsumeRecordStream()
             {
                 var result = ResultCreator.CreateResult(1, 2);
 
                 result.Dispose();
 
                 result.AtEnd.Should().BeTrue();
-                result.Position.Should().Be(2);
+                result.Position.Should().Be(-1);
                 result.GetEnumerator().MoveNext().Should().BeFalse();
                 result.GetEnumerator().Current.Should().BeNull();
             }
 
             [Fact]
-            public void ShouldPullSummary()
+            public void ShouldNotPullSummary()
             {
                 int getSummaryCalled = 0;
                 var result = ResultCreator.CreateResult(1, 0, () => { getSummaryCalled++; return new FakeSummary(); });
 
                 result.Dispose();
-                getSummaryCalled.Should().Be(1);
+                getSummaryCalled.Should().Be(0);
+            }
+
+            [Fact]
+            public void PeakShouldThrowAfterDispose()
+            {
+                var result = ResultCreator.CreateResult(1, 2);
+                result.Dispose();
+
+                Assert.Throws<ObjectDisposedException>(() => result.Peek());
+            }
+
+            [Fact]
+            public void ConsumeShouldThrowAfterDispose()
+            {
+                var result = ResultCreator.CreateResult(1, 2);
+                result.Dispose();
+
+                Assert.Throws<ObjectDisposedException>(() => result.Consume());
+            }
+
+            [Fact]
+            public void SingleShouldThrowAfterDispose()
+            {
+                var result = ResultCreator.CreateResult(1, 2);
+                result.Dispose();
+
+                Assert.Throws<ObjectDisposedException>(() => result.Single());
             }
         }
 
--- a/Neo4j.Driver/Neo4j.Driver/Driver.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Driver.cs
@@ -73,7 +73,7 @@ namespace Neo4j.Driver
         ///     Establish a session with Neo4j instance
         /// </summary>
         /// <returns>
-        ///     An <see cref="ISession" /> that could be used to <see cref="ISession.Run" /> a statement or begin a
+        ///     An <see cref="ISession" /> that could be used to <see cref="IStatementRunner.Run(Statement)" /> a statement or begin a
         ///     transaction
         /// </returns>
         public ISession Session()
--- a/Neo4j.Driver/Neo4j.Driver/Driver.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Driver.cs
@@ -59,8 +59,13 @@ namespace Neo4j.Driver
         {
             if (!isDisposing)
                 return;
-            _sessionPool?.Dispose();
-            _sessionPool = null;
+
+            if (_sessionPool != null)
+            {
+                _sessionPool.Dispose();
+                _sessionPool = null;
+            }
+
             Logger?.Dispose();
         }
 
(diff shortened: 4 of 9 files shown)


Neo4j.Driver/Neo4j.Driver.Tests/Result/PeekingEnumeratorTests.cs
```diff
@@ -115,15 +115,32 @@ public void ShouldConsumeAllItemsAfterDiscard()
...
-                enumerator.Consume();
+                enumerator.Dispose();
                 enumerator.MoveNext().Should().BeFalse();
                 enumerator.Peek().Should().BeNull();
-                enumerator.Position.Should().Be(3);
+                enumerator.Position.Should().Be(-1);
+            }
+
+            [Fact]
+            public void PeakShouldReturnDefaultAfterDispose()
```
zhenlineo, line 129: I do not think `Peek` should return something different from `Current`
The `Current` method will return `null` right now, see code [here](https://github.com/neo4j/neo4j-dotnet-driver/blob/1.1/Neo4j.Driver/Neo4j.Driver/Internal/Result/PeekingEnumerator.cs#L58)

Right now we do not have the problem of using `null` or `default(T)` as we actually use `list<object>`. But if we changed to `default(T)`, then I was worried that we might have some problems for `int` if we changed to `list<int>` or `list<dynamic>` or `list<T>` in the future. If we still use `null`, at least we got the option to use `int?`. 
But if we return 0, then it might result in a strange bug in the end.

IngvarKofoed, line 129: I agree. Peak should return default(T) instead of null in the code: `return CacheNext() ? _cached : null;`

Sorry, I missed that one..

IngvarKofoed, line 129: Sorry, I think my changeset is ok. I have made the code act the same way as other enumerators in the standard .NET framework. They either throws or return frfault values. 

I think that divegating from standards can lead to bugs. 

We can leave out this commit and fix thr warning another way?

zhenlineo:
There are some rules when designing this result API:
- The close of a result (`result.Dispose`) should mark the end of a statement conversation(running a statement and get all result back, etc.) with the server.
- The records in the result could only be visited sequentially until the `result.Despose` is called.

The first rule implies that `result.Dispose` has to `Consume` all records as that's the only way to force the query to finish, a.k.a. all records are pulled back from the server, a.k.a. all data in packstream buffer are consumed and the buffer is ready to start a new fresh session.

The second rule implies that the following code should work:
The following code should work with this result API:

```
            [Fact]
            public void ShouldConsumeRecordsSequentially()
            {
                var result = ResultCreator.CreateResult(5); // say we have [0, 1, 2, 3, 4] as records in result
                result.Take(2).ToList(); // expecting to get [0, 1] back, PeekingEnumerator.Dispose is called
                result.Position.Should().Be(1);
                result.AtEnd.Should().BeFalse();

                result.Take(3).ToList(); // expecting to get [2, 3, 4], which requires the previous Dispose should do nothing otherwise this one will be affected.
                result.Position.Should().Be(4); // with your code we got 1 here currently
                result.AtEnd.Should().BeTrue();
            }
```

Your current changes in the PeekingEnumberator and StatementResult would break the contract now. But if there is other options to solve this, I am open to change it to be more C# or dotnet style. 
*\* also pls let me know if I misused `Take` method.

Neo4j.Driver/Neo4j.Driver.Tests/Result/PeekingEnumeratorTests.cs
zhenlineo, line 129: I would suggest in this PR we just focus on fixing the compiling warning, and then leave the changes to the result API to another PR, as we currently have disagreement on the use of Dispose and enumerator, WDYT?

IngvarKofoed:
Ok, so let me write in my own words, what I have read from you comment :)

API rule 1:
The PeekingEnumerator has to read the rest (all) because the enumerator it is given in its constructor has to finish to pull the rest of the records from the Neo4j server? In that case, my change is breaking that. I will make changes to fix this. But I think you and your colleagues should rethink this strategy, because it is not possible for a client to 'bail out' early and is forces to pull all records. So code like this will have a less optimal performance:

```
using(var driver = GraphDatabase.Driver( "bolt://localhost:7687" ))
using(var session = driver.Session())
{
    var result = session.Run("CREATE (n) RETURN n");
    foreach (var record in result)
    {
        if (something) 
       {
           return; 
           // Bail out early and as I client I do not care about the rest of the records
           // and do not want to spend time consuming them
       }
    }
}
```

API rule 2
It seems that there is a bug in both in the existing and in my version of StatementResult. The method public IEnumerator<IRecord> GetEnumerator() should always return a new enumerable object. This is a .NET "requirement" of IEnumerable<T>. See https://msdn.microsoft.com/en-us/library/s793z9y2(v=vs.110).aspx 
I think in general, there is an issue with the PeekingEnumerator and the use of it in StatementResult.  I do not think this is how one would normally program this in C#.

Ill happily create a PR for both these, if you like? :)

zhenlineo:
You got all my ideas correctly.
And I always have the feeling that the PeekingEnumberator is impled wrongly :P

For the performance problem and also with the issue https://github.com/neo4j/neo4j-dotnet-driver/issues/39, I feel it might be a bad idea to put the `Consume` inside the `Dispose`. But still we need to provide user a way to mark the end of one "conversation" with the server. If I do not put `Consume` indside `Dispose`, then the use would be changed to

```
using(var driver = GraphDatabase.Driver( "bolt://localhost:7687" ))
using(var session = driver.Session())
{
    var result = session.Run("CREATE (n) RETURN n");
    foreach (var record in result)
    {
        if (something) 
        {
           break; 
           // could only break but not return
        }
    }
    result.Consume() // force to get all records back if we ever break in the middle of a loop.
}
```

I am not sure if this is an improvement though. The code rely mostly on a user to remember to call `Consume()`.

However, an argument against your performance problem is: Do not pull the results if you do not need them! Instead you should filter with Cypher rather than pull everything to the client! :)

lassewesth:
If you make Session do result.Consume() as it's Dispose behaviour now, then we can later change it so that it does result.AwesomeConsume() - without asking clients to update their code

IngvarKofoed:
@zhenlineo Regarding the Take method (and other methods from Enumerable). Yep, you have misunderstod it :) Take and the other methods are pure meaning that they do not alter the collection that they are called on. They return a new IEnumerable object that will yeild the result of the method. Just like string.Trim() does not change the value of the string, but returns a new string instance. 

Explained in code:
var result = new List<int> { 1, 2, 3, 4, 5 };
var a1 = result.Take(2).ToList(); // a1 is a List<int> that contains: 1, 2
var a2 = result.Take(3).ToList(); // a2 is a List<int> that contains: 1, 2, 3

Hope this explains it :)

zhenlineo:
Thank for the explanation. But the idea is a single direct cursor - that is you cannot move backed once you have visited the result. I know it breaks the Take method, but we disallow a user to revisit the records that already moved over.

var result = new MySingleDirectionList{ 1, 2, 3, 4, 5 };
var a1 = result.Take(2).ToList(); // a1 is a List that contains: 1, 2
var a2 = result.Take(3).ToList(); // surprise: a2 is a List that contains: 3, 4, 5 because 1, 2 is already gone

Here is the dilemma:
I design my own methods to do this single direction cursor in IStatementResult and lost the ability to use the result in a foreach loop
Or
I use enumerable and break all assumptions a normal developer would expect :(

jakewins:
@IngvarKofoed do you have a reference for Take and friends being side-effect free? The docs just say `Take` "Returns a specified number of contiguous elements from the start of a sequence.".

As it stands, IEnumerable seems to be the standard for representing blocking stream APIs (as highlighted on the [Rx.NET intro page](https://github.com/Reactive-Extensions/Rx.NET), for instance) - meaning there's good precedence for using Enumerable to represent a blocking stream API as we are doing here. We can't make the API immutable, because the result streams can be infinite or, if finite, in the hundreds of gigabytes range, implicitly making copies of the stream is not an option :(

IngvarKofoed:
@jakewins `Take` and friends are all extension methods taking `IEnumerable<T>` as the first argument. See the remark section of the `Enumerable` class that contains all the methods here: https://msdn.microsoft.com/en-us/library/system.linq.enumerable(v=vs.100).aspx. If you combine this remark will the fact that `IEnumerable<T>` is for enumerating and has no methods for changing the underlaying collection (Ex: List<T>) I think that "proves" that they do not have side effects. But `ToList` does (see below).  

I do not see any issue by using `IEnumerable<T>` as you do - to represent a blocking stream. My fix were for the Dispose which I still think is a very valid fix (Though the fix broke something - which is bad). The issue is that when you call the method `ToList` (as the unittest does), it will call Dispose on your `IEnumerator<T>` returned by the `IEnumerable<T>` because ToList iterates through all the elements and adds them to the list object. For what you are describing, calling `ToList` would be bad since it could be gigabytes of data or never terminate. 

I have an idea of fixing both the Dispose method and and still getting what you and @zhenlineo describes, but done in a more C#/.NET fashion that behaves even more like what C# developers expects :)

I can create a PR for it, but Ill might have to change the unittest to not use ToList because of what I have just described. What do you think?

IngvarKofoed:
Just a little update: I have working solution that passes all unittests. Its a very simple solution that seems to fit your needs :) I only had to make minor changes to a couple of unittests regarding the `Position` and I think I can fix that also - off-by-one issue.

I just need to understand the idea and motivation behind `Single` and `Peek` methods? 
Why have `Single` method, when there is an `Enumerable.Single` method?
What is the use case for the `Peek` method? Why not just let the user use methods like `Enumerable.First` or `Enumerable.FirstOrDefault`?

jakewins:
Nice! Thanks for helping out with this btw.

For peek, there are several use cases - the most prominent of which is debugging and other "side channel" operations where you're handling the stream already in a pipeline, but want a way to patch into the stream without affecting its state:

```
System.Console.WriteLine( "About to do stuff with:" + result.Peek() );
DoStuffWith( result );
```

There are also a few use cases around dispatch, where you may want to choose which method to dispatch to depending on the next record in the stream, again without actually changing the state of the stream.

For `Single` - this is different from `First`, `First` is useful when you want one record, and don't care if there were lots more found. `Single` is when you want a single result, and it is an error if there are more records. `Single` is rather common in database scenarios where we often deal with unique entities, like `User` or whatever. If you ask the database for a user by email, say, it's nice to use `Single` instead of `First` to guard against domain corruption and security breaches - if you ever got more than one user back from that statement, it's better to fail than to guess that the first one is the correct user entity.

IngvarKofoed:
Thanks!! Sure, I love to do look at code at and try to improve it :) I hope to have a PR ready later today

Alright, now I under stand `Peek` :)

Regarding `Single`, please reread what I wrote and also have a look at link below, because I was not talking about `First` or `FirstOrDefault` but about the build in `Single` method in the .NET framework :) So if it is for what you describe, I do not really see a need for a custom `Single` method. But if you really want a `Single` method, could you live with that it will that the `IStatementResult.Single` is implemented using the build ind 'Single' meaning that the exceptions will be build in and not custom? Im asking because it feels strange to implement something with multiple lines and several unittests, that could be a single-liner or even not in the driver code :)

https://msdn.microsoft.com/en-us/library/bb155325(v=vs.100).aspx

IngvarKofoed:
Ah, now I see where `IStatementResult.Single` gives more than the build-in `Single`. `IStatementResult.Single` also should handle that you have not already read n-1 of a total n records. In this case the build-in `Single`does not fail :)

IngvarKofoed:
Update: My strategy seems to be working very good! I have also implemented Single and Peek and did not have to change any unittests :) Im not ready to make a PR because I still need to do some minor things. Spend too much time on discovering that the integration tests failed because the Neo4j server nightly build was updated and it was not my changes that made the tests fail.

Hoping for a nice PR tomorrow :)

IngvarKofoed:
You can see working code here: https://github.com/IngvarKofoed/neo4j-dotnet-driver/tree/1.0-enumerable-fix-for-statement-result

I will clean it up and make it ready for a PR. But not sure if it should be part of this PR or not?

zhenlineo:
I moved the non-controversial part into #46. If no one objects, I will merge #46 first.
I will keep this PR in case anyone wants to continue to discuss until we get a new place for discussion.

IngvarKofoed:
Sounds good with #46 I had the same thought :)

IngvarKofoed:
I have created a PR that covers the discussion within this PR about the StatementResult, consuming and forward only iteration :) See PR #50

zhenlineo:
Decision made in https://github.com/neo4j/neo4j-dotnet-driver/pull/50

PR #48, opened 2016-04-01
1.0 fix warning for statement result dispose

zhenlineo:
The IDisposible is added on the impl StatementResult
The user should not do

```
using(Driver driver = GraphDatabase.Driver(url))
using(ISession session = driver.Session())
using(IStatementResult result = session.Run("A Cypher statement"))
{
    //do sth with the result
}
```

Instead, the correct way of use this result API should be

```
using(Driver driver = GraphDatabase.Driver(url))
using(ISession session = driver.Session())
{
    IStatementResult result = session.Run("A Cypher statement");
    //do sth with the result
    result.Consume(); // explicitly dispose all resources in result.
}
```

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Result/PeekingEnumerator.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Result/PeekingEnumerator.cs
@@ -107,6 +107,7 @@ namespace Neo4j.Driver.Internal.Result
             _position ++;
             _cached = null;
             _current = null;
+            _enumerator.Dispose();
             _enumerator = null;
             _hasConsumed = true;
         }
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Result/StatementResult.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Result/StatementResult.cs
@@ -25,10 +25,10 @@ namespace Neo4j.Driver.Internal.Result
     /// <summary>
     /// The result returned from the Neo4j instance
     /// </summary>
-    public class StatementResult : IStatementResult
+    public class StatementResult : IStatementResult, IDisposable
     {
         private IResultSummary _summary;
-        private readonly IPeekingEnumerator<Record> _enumerator;
+        private IPeekingEnumerator<Record> _enumerator;
         private List<string> _keys;
         private Func<IResultSummary> _getSummary;
         internal long Position => _enumerator.Position;
@@ -106,6 +106,10 @@ namespace Neo4j.Driver.Internal.Result
                 return;
             }
             Consume();
+            if (_enumerator != null)
+            {
+                _enumerator.Dispose();
+            }
         }
 
         public void Dispose()


IngvarKofoed:
Im sorry to say, but it does not make sense to add `IDisposable` to the `StatementResult` implementation because the `Dispose` method will never be called.

Also asking the users to call the `Consume` method is bad practice in C#/.NET. If something needs to be done/called after you are done with an object, it should implement `IDisposable`. In this case it should be the `IStatementResult` interface that implements `IDisposable`. But doing this will require the user to do what you write they should not: 

```
using(Driver driver = GraphDatabase.Driver(url))
using(ISession session = driver.Session())
using(IStatementResult result = session.Run("A Cypher statement"))
{
    //do sth with the result
}
```

So I think you should rethink https://github.com/neo4j/neo4j-dotnet-driver/pull/48/commits/15373f116e1504283f2b2cd4d845d67e9ef14347. I suggest that  `IStatementResult` interface implements `IDisposable` and you use my fix for the warning in (when its done): https://github.com/IngvarKofoed/neo4j-dotnet-driver/tree/1.0-enumerable-fix-for-statement-result

zhenlineo:
@IngvarKofoed I am actually confused what is the thing that need to be disposed explicitly. For my understanding, the Dispose method is not needed if all the records in the result list are consumed. I actually could just leave this Dispose method out as I have nothing to dispose at this moment.

Say if we have a code

```
IStatementResult result = session.Run("A Cypher statement");
for(IRecord record : result) // If I visit all the records in this result
{
// whatever
}
```

Then there is no problem without calling a Dispose, as there is no special resource need to be released. Am I correct here?
The only thing is `var enumerator = recordList.GetEnumerator()`. This enumerator is not disposed. But I thought they did nothing inside this enumerator.Dispose().

The only time that a user need to call Consume is that the user does not finished his iteration.
Would you explain me what resource I need to release right now?
We assume that the user should always consume the records, either by visiting each of them or by calling Consume to throw them away. So there is no need to have a Dispose method for this case, isn't it?

IngvarKofoed:
Ill try to explain my thought process, so here goes :)

There are some issues with the code: 

`var enumerator = recordList.GetEnumerator()`

First: It is not something that is normally done in C#. You will use `foreach`, or some of the `Enumerable` extensions or `ToList`. And all of these will call dispose on `enumerator`. This means that the `Dispose` will be called 0 or more times on the enumerator object returned by `StatementResult.GetEnumerator()`. This implies that `StatementResult.GetEnumerator()` has to return a new object every time it is called.

Second: The `Consume`needs to be called on `StatementResult` because, what if the user never calls, either indirect or direct the `GetEnumerator()`. Then the records never gets consumed. 

And this leads to that `IStatementResult` needs to implemented `IDisposable` to follow standard C# patterns.

So as I see it, to be standard C#, there are two options:

1) `IStatementResult` implementens `IDisposable` and uses the implementetion given here or something similar where `StatementResult.GetEnumerator()` has to return a new object every time it is called.

2) You have to drop the consume requirement.

Did this explain it? If not, please let me know and Ill give it another try :)

zhenlineo:
decided in https://github.com/neo4j/neo4j-dotnet-driver/pull/50

PR #49, opened 2016-04-02
Disposable fix for StatementResult

IngvarKofoed:
This PR fixes the warning that was in the StatementResult. It keeps true to the API of both the Neo4j server and the driver API by consuming all records regardless if the user needs them and that only forward (and no restart) iteration is possible - eg consumes records when enumerated. 

The only "breaking" thing is that `IStatementResult` has to be `IDisposable` to have the consume regardless feature. This is the correct way to do it in C#/.NET. I do not see this as a problem, because there might be other reasons in the future that ÌStatementResult`needs to be disposed and same behaviour is seen in other APIs for DBs.


IngvarKofoed:
Wrong branch, should have been on 1.0

PR #50, opened 2016-04-02
Disposable fix for statement result

IngvarKofoed:
This PR fixes the warning that was in the StatementResult. It keeps true to the API of both the Neo4j server and the driver API by consuming all records regardless if the user needs them and that only forward (and no restart) iteration is possible - eg consumes records when enumerated.

The only "breaking" thing is that IStatementResult has to be IDisposable to have the consume regardless feature. This is the correct way to do it in C#/.NET. I do not see this as a problem, because there might be other reasons in the future that ÌStatementResult`needs to be disposed and same behaviour is seen in other APIs for DBs.


zhenlineo:
Thanks so much. I talked with Chris and my collea a lot about this API change. 
We would like to accept your changes to the enumerator, where GetEnumerator always returns a new enumerator and the enumerator get disposed properly when it is done. Your impl for this part is really neat.

But we still disagree with the the dispose method added on `IStatementResult` (though in this PR you actually did not make any changes to the interface but the impl)
So the idea is that the driver provides two ways of consuming the result:
1. lazy consumption
2. eager consumption

The Dispose method you added is actually helping us to impl eager consumption of the records in a result. So the Dispose method will eagerly pull all the records and block until the final record comes.

Without (Consume called in) Dispose method, the records are lazily pulled.

```
using(var session = driver.session())
{
    var result1 = session.Run("something");
    // if result1= [1,2,3], and at this moment, we only got [1]
    var result2 = session.Run("another thing");
   // while we sending run st, the result1 got [1.2.3]
   result2.Consume(); // this will wait for result 1 to receive [1,2,3], and all records in result2
}
```

As you could see in this example, we could choose not wait for result1. Then we could modify underlying impl to do batching for records for result1 and result2 or other network actions.

So all in all, as we still have so many opinions on this Dispose method, how about we leave it for now and add it when required?

zhenlineo:
Pls let me know if you have any concern about not having a Dispose method on `StatementResult` and/or `IStatementResult`.

zhenlineo:
If you agree to remove Dispose, I get this PR
https://github.com/IngvarKofoed/neo4j-dotnet-driver/pull/3

IngvarKofoed:
It is fine with me to not have the `IDisposable` in the `IStatementResult`. I just thought it was a API requirement (API that the Neo4j server exposes) to consume the rest of the records: See https://github.com/neo4j/neo4j-dotnet-driver/pull/42#issuecomment-203389162. And in https://github.com/neo4j/neo4j-dotnet-driver/pull/47 you have this code: 

```
using(Driver driver = GraphDatabase.Driver(url))
using(ISession session = driver.Session())
{
    IStatementResult result = session.Run("A Cypher statement");
    //do sth with the result
    result.Consume(); // explicitly dispose all resources in result.
}
```

And like I wrote in that PR, it is not best practice to require users to do a `result.Consume();`. Best practice would be to do the following: 

```
using(Driver driver = GraphDatabase.Driver(url))
using(ISession session = driver.Session())
{
    using(IStatementResult result = session.Run("A Cypher statement"))
    {
        //do sth with the result
    }
    using(IStatementResult result = session.Run("A nother Cypher statement"))
    {
        //do sth with the result
    }
}
```

Which requires `IStatementResult` to implemented `IDisposable`.

But as I said, its up to you :) So please let me know if you really want to remove the `IDisposable` from the `IStatementResult`. If you do, Ill merge your PR so it is ready :)

IngvarKofoed:
I have just reread you code example. I find it hard to see how you would accomplish this unless you put the 'batching' stuff in the session. See the NEW comments in your code below:

```
using(var session = driver.session())
{
    var result1 = session.Run("something");
    // if result1= [1,2,3], and at this moment, we only got [1]
    var result2 = session.Run("another thing"); // NEW: What if result1 is not done at this point? Would that not break the API rule to the Neo4j Server, because the server is not finished sending result1 records before you make a new query?
   // while we sending run st, the result1 got [1.2.3]
   result2.Consume(); // this will wait for result 1 to receive [1,2,3], and all records in result2
  // NEW: How would result2 now that it has to consume the rest of result1? I guess you would store this information in the session. If that is true, why not just do the consuming in the session Dispose method instead of calling Consume on result2.
}
```

It seems a little strange to me that calling `Consume` on `result2`, also consumes `result1` records. What if the user did something like this:

```
using(var session = driver.session())
{
    var result1 = session.Run("something");
    var result2 = session.Run("another thing"); 
    result.Consume();
    foreach (var record in result1)
    {
        // Do someting with record
    }
}
```

I guess this would fail? I think your idea is leaking the 'dirty' details of how the driver must communicate with the Neo4j Server out to how to use the driver (consuming of records). I think this is a bad design and it will make it hard to change in the future. 

I would recommend to remove the `Consume` method from the public part of the driver API. This is a 'dirty' detail that users of the driver should not be bothered with and it is something that I hope the Neo4j server is going to change in the future. And as I have suggested, use the common way of doing cleanup in .NET by using the `IDisposable` pattern :)

zhenlineo:
We have this test:
https://github.com/neo4j/neo4j-dotnet-driver/blob/1.1/Neo4j.Driver/Neo4j.Driver.IntegrationTests/ConnectionIT.cs#L163

zhenlineo:
The consume still just consume the records in the result2 as it is a enumerator created from the records on result2. result1 get all its records buffered. 
The problem is that result1 and result2 is sharing the same connection to stream the result back.

```
output channel:
run st1
run st2
```

```
input channel:
record1 for st1
record2 for st1
...
recordn for st1
summary for st1
record1 for st2
record2 for st2
...
```

As you could see, the result2 could not be proceed if result1 is not stream back. However consume the record on each result is separate

IngvarKofoed:
Ok, so according to that test, the order of which you consume all the records is not important. Did I understand this correct? 

How would the test look like if you did nothing to `result1`? Example:

```
using (var driver = GraphDatabase.Driver("bolt://localhost"))
using (var session = driver.Session())
{
    var result1 = session.Run("unwind range(1,3) as n RETURN n");
    var result2 = session.Run("unwind range(4,6) as n RETURN n");

    var result2All = result2.ToList();

    result2All.Select(r => r.Values["n"].As<int>()).Should().ContainInOrder(4, 5, 6);
    // Should there be a result1.Consume or result2.Consume here?
}
```

IngvarKofoed:
It seems to me that the session should guard agains running a new statement before the first one is done? Because you will not get the result of the second before the first is done, right?

zhenlineo:
You are right. That's why we have a `Consume` method to guard a clean state if we did not visit all of the records and/or pull summary.
if we have
`result2.ToList()`
Then it is guarded that all records for result1 and result2 is back.
if we have
`result2.summary`
Then it is guarded that the state is clean for new statement.
We only need to call `Consume` if we did not visit all record and we want to make sure the state is clean for next statement.
However it is better to let the user to decide (by calling `Consume`), rather than always clean the state for them.

IngvarKofoed:
My point is that the user should not even know about cleaning the state, this is an internal matter. 

I like the idea that the unittest implies that, that it happens without the users knowledge and that it does not require the `Dispose` method. So Ill merge your PR :)

I think you should consider removing the `Consume` method from the public API because I see no reason for it. The consuming happens when it is needed (lazy or eager).

zhenlineo:
The consume is mainly there for uniformity of all drivers. We do not need `Consume`, `Single`, `Peek` in .Net probably as every .Net developer will just do a foreach loop directly on result (by Chris) :D

Thanks a lot for helping building this PR again!

IngvarKofoed:
Ah, I see, I properly misunderstod you comment in #47 that it was required by the users to call the `Consume` method. Sorry about that :)

No problem, I enjoy this and have a lot more that I think I can contribute with :)

Issue #40, opened 2016-03-27
All public types should be documented


IngvarKofoed:
When #38 has been resolved, then all public types should be documented to make the driver more user friendly and help developers not make less mistakes and use lesser time on getting starting with the driver.

zhenlineo:
Oh, I replied in #38, all the classes/interfaces that we would like the user to use actually are doc properly (or at least with docs). 
The dotnet API doc could be found [here](http://alpha.neohq.net/docs/dotnet-driver/html/f7d9eea1-5357-f1de-8a54-336a77141b6d.htm) in [alpha channel](http://alpha.neohq.net) 
Also the API doc will also be moved to a public/formal place soon.

Let's revisit this issue when we solved #38.

zhenlineo:
The docs are updated with all public class documented now [dotnet-api-doc](http://alpha.neohq.net/docs/dotnet-driver)

zhenlineo:
Close this one as it is already resolved now

PR #53, opened 2016-04-05
Making official nonpublic types internal

IngvarKofoed:
All types that are not listed here: http://alpha.neohq.net/docs/dotnet-driver/html/f7d9eea1-5357-f1de-8a54-336a77141b6d.htm
was made internal with the internal keyword. 

I also moved the Neo4j exceptions to root namespace because it makes the Neo4jDriver easier to use by clients because they do not have to include two namespaces. I see the exceptions as common types and therefor they should be en root namespace. This is the norm for the standard C#/.NET framework. 

This is a "fix" for issue #40


zhenlineo:
WDYT https://github.com/neo4j/neo4j-dotnet-driver/pull/54?

PR #54, opened 2016-04-06
1.0 internal types

zhenlineo:
Based on https://github.com/neo4j/neo4j-dotnet-driver/pull/53
Rebased the PR and made LoggerBase internal too so that the public API and internal uses have a clear cut now.


IngvarKofoed:
I like very much that you have made the change with the `LoggerBase` and removing it from the `Driver`! :)

I think this is a very good merge :)

PR #58, opened 2016-04-10
Bug fix for streaming records

zhenlineo:
Fix the bug where the pos/index in record stream is not moving forward correctly.
Fix the bug where the peek method starts from record 1 rather than 0.

Before the fix, some Linq methods will not consume the record when the record has been visited.
This is very wrong as it enables the user to buffer the pos in the stream and revisit it latter.
The bug is caused by not forwarding the current pos/index in the stream before yielding the record back.

The peek method now would peek the record that will be returned by calling Enumerator.MoveNext()
which means it will return record in pos 0 before Enumerator.MoveNext() is called the first time.


technige:
:+1:

PR #61, opened 2016-04-12
Making an interface for the Driver 

IngvarKofoed:
Reasons for doing this:
- The code is more complete, every type used except the `GraphDatabase` is an interface
- It is easier to make code that can pass an `IDriver` instance to a class and then make unittests for this class.
- In the future, it could be possible to have different driver implementations that the `GraphDatabase` returns depending on settings etc. - without breaking existing client code.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/Examples.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/Examples.cs
@@ -290,7 +290,7 @@ namespace Neo4j.Driver.Examples
 
         private void ClearDatabase()
         {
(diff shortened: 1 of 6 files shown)


zhenlineo:
It is a reasonable change. I will try to look into the red test build tomorrow. It seems a server problem to me. I thought the build will recover itself, but I 'd better double check it tomorrow.

zhenlineo:
I was super confused by the test failure in `DriverAuthSteps`. It turns out the bug is we forget to change to `IDriver` there. Then I rebase the PR and get this one https://github.com/neo4j/neo4j-dotnet-driver/pull/63.

PR #66, opened 2016-04-20
Moved all API into `Neo4j.Driver.V1`

zhenlineo:
By naming it be V1, we could achieve supporting V1, VX inside the same codebase.


IngvarKofoed:
I think this is a bad move. I have never seen this strategy for versioning used in any of the many 3rd party C# libraries I have used. 

The strategy I have seen in practically all cases, is that there is not introduced any breaking changes in public types unless the major version is increased. In case where the major version is increased, the release notes states that there is breaking changes and as a developer I can see if I want to use them or not. 

With this suggested strategy, if I want to upgrade, I have to go through all my source files and make sure that .V1 is not used, but .VN is used instead with the danger of missing one of them and then my code might use two different version of the driver at the same time.

And in all, I would strongly advice you not to use this versioning strategy! :)

technige:
Thank you for your thoughts on this. I completely understand that this may not be your preferred approach but after a lot of internal discussion, we decided that we wanted to be explicit about the API version within package names. This change is to bring the .NET driver in line with the other drivers.

Part of the reason for this decision is that we want to draw the API definitions out into a separate specification later on, in conjunction with the TCK. A core set of methods and properties that are uniform across the drivers will themselves be versioned and this namespace highlights that explicitly.

IngvarKofoed:
Yes, I guessed this was the reason :)

My final word on this is, that this makes both the public and private parts of the C# driver look clumsy and foreign for C# developers. Something I have noticed in some other parts of the driver implementation. This, at least in my opinion, could affect the adoption rate of the driver within the C# community and that's why I wrote my first comment on the PR.

But I fully respect you choice, just wanted to let you know :)

boggle:
We do not expect to bump this version number in a very long time, hence it is quite stable to put this into source files.   In fact the only point in time for that will be if we need to radically change the Driver API and hence must publish a driver that will support two major API versions in one release for application migration purposes.

IngvarKofoed:
@boggle Ok, it then seems to me that the .V1 namespace postfix is not needed.... :)

boggle:
@IngvarKofoed what I thought as well until I realized that for migration one needs two APIs in one driver and if you don't version them that will be really confusing.. :)

IngvarKofoed:
@boggle Yes, I see and understand the migration use case :) 

My take on it is that migration is a "corner case" use of the driver and all other use cases which, I guess, would be the large majority does not care about this. So making the majority "pay" for the single use case seems a little overkill to me :)

boggle:
Fair enough.

lassewesth:
@boggle I can't see why you would want to have 2x driver versions in your app at the same time.

For migration, the way I have done it and seen it done is, you substitute-dependency-and-fix. You _never_ stop at the half way house where you have migrated half your app.

Are you thinking of a case where several apps share a soup of dependencies? At that point your problems are way bigger than namespaces...

boggle:
I'm thinking of a large application that you migrate piece by piece to a new version of an API.

boggle:
It also gives some clarity in the code, you see which version of the API is used.

jakewins:
@lassewesth this happens constantly in large systems. It's the problem OSGi tried to solve, and it's what Oracle is trying to solve with Jigsaw. If you depend on Lucene, and pull in a project that also depends on Lucene (ahem..), because of the way the Lucene project does namespacing, you can't do this unless you depend on the same versions of Lucene. Same problem pops up _constantly_ in Scala projects, for instance.

Contrast that to Netty, where you can easily have one part of your system using some hipster HTTP stack that is based on Netty 3, and a REST API in your system that's using Netty4, because they segregate the name spaces between those two major versions.

Using this naming scheme, the same thing will be true for Neo4j - if you use Neo4j for your main app, and then you pull in `Honest Joes Metrics Storage` that happens to also use a Neo4j client, you won't have version clashes with the `Honest Joes` dependency if you use the Neo4j Driver.

Eg; this solves a huge pain that pops up over and over, and the only counter argument I've heard is that it looks wierd, which IMO does not override the rather tangible benefit this gives.

IngvarKofoed:
There is more elegant ways then having multiple of having side-by-side versions of the same assembly in .NET, than having multiple versions in the code. You can use either the GAC, that requires that the assemblies are strong named or you can use assembly redirect in the app.config. 

It seems to me that you are making it easy for enterprises and making the code harder and error prone for entry and mid levels projects. With the danger of scaring new customers away. And it also seems that, if you do not know about the single usecase of migration, that Neo4j did not bother to handle versioning and left it up to the user to fingure things out. 

As a trendsetting developer for 30+ developers at my job, I would seriously reconsider using any 3rd party library with this versioning schema. Because I would have to tell all developers to use the namespace .V2, not .V1 or .VN. Which can be hard because VS will pop all choices in the quick-fix-menu. Then I would have to cross my fingers that all developers, including newly started, remembers this or rely on that the compiler or automatic tests will catch any wrongly used namespaces. Which might not be the case if you have somekind of runtime loading of plugins using MEF or similar. 

I would much rather use NuGets or similar to handle my versioning of 3rd party tools, this is one of the primary tasks of NuGets. And in the off-case where I have to handle migration Ill use the build-in tools in the .NET runtime for this (GAG and assembly redirect). 

I'm not writing all this just to "be right". I'm writing this because in all my many years of C#/.NET development, I have almost never seen this version strategy. And it will be harder for me to use this driver professionally, which makes me sad because I'm constantly looking for at good case to introduce Neo4j at my work :)

jakewins:
@IngvarKofoed your feedback is well heard. I don't think we will change our minds on this topic for the `1.0.0`, partially because this has been up for discussion several times around already (and we feel reasonably comfortable with it still given that background) and we're very close to GA, but please know that your insights here and elsewhere has been incredibly appreciated and has sparked lots of good discussions internally as well. 

One note, of course, is that this is intended as a baseline; there's nothing stopping projects using this as a fast driver to sit on top off and to explore alternative API approaches!

IngvarKofoed:
@jakewins Alright :) Happy to hear that I can contribute with something and I myself also get something out of having these discussions :)

PR #92, opened 2016-06-29
1.0 streaming

zhenlineo:
**Example 1**:
If I run

```
using(var session = driver.Session())
{
    var result = session.Run("RETURN 1");
}
```

The sequence of messages to exchange between client and server are:

```
-> INIT
<- SUCC_init

-> RUN
-> PULL_ALL
<- SUCC_run
```

After `}` the `session.Dispose()` will result in message sequence:

```
-> RESET
<- RECORD 1
<- SUCC_pull_all
<- SUCC_reset
```

So the RUN, PULL_ALL messages are eagerly sent, the SUCC for RUN is also eagerly received, but for records, it will only be retrieved in `result.Consume()` or similar iterations. If no result is iterated before the session is disposed, then the `RESET` message will force to sync all unhandled messages to make sure the session is clean when it returned to pool.

**Example 2**:
When I run

```
using(var session = driver.Session())
{
    session.Run("RETURN 1");
    var result = session.Run("RETURN 2");
    result.ToList();
}
```

Then the sequence of the messages are:

```
-> INIT
<- SUCC_init

 // run(return 1)
-> RUN
-> PULL_ALL
<- SUCC_run

// run(return 2)
-> RUN'
-> PULL_ALL'
<- RECORD 1
<- SUCC_pull_all
<- SUCC_run'

// result.ToList
<- RECORD 2
<-SUCC_pull_all'
```

In this case, as we did not save the result of the first `Run`, so they will just get lost if the second `Run` start to consume the result first. (We will improve this to cleverly buffer the results latter)

**Example 3**:
If we have

```
using(var session = driver.Session())
{
    try{session.Run("Invalid");}
    catch{}
    session.Run("RETURN 1");
}
```

Then the sequence of messages for the first `Run` are:

```
-> INIT
<- SUCC_init

-> RUN
-> PULL_ALL
<- FAIL_run
<- IGNORE_pull_all
-> ACK_FAIL
```

Note, as we failed before streaming, on the `RUN` message, we will then give up streaming but just drain all the messages that wait for the reply immediately. Therefore we will also wait for `IGNORE` message for `PULL_ALL` message. Finally the `ACK_FAIL` is added so that we will try to reset the session back to a good state for the following statement. 

Then the messages for the second `Run` is the very same as the first example:

```
<- SUCC_ack_fail
-> RUN
-> PULL_ALL
<- SUCC_run

-> RESET
<- RECORD 1
<- SUCC_pull_all
<- SUCC_reset
```

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/MessageResponseHandler.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/MessageResponseHandler.cs
@@ -62,6 +62,7 @@ namespace Neo4j.Driver.Internal.Connector
                 // before summary method is called
                 CurrentResultBuilder?.CollectSummaryMeta(meta);
             }
+            Error = null;
             _logger?.Debug("S: ", new SuccessMessage(meta));
         }
 
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/MessageResponseHandler.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/MessageResponseHandler.cs
@@ -14,6 +14,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+using System;
 using System.Collections.Generic;
 using Neo4j.Driver.Internal.Messaging;
 using Neo4j.Driver.Internal.Result;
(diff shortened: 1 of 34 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/MessageResponseHandler.cs
```diff
@@ -83,35 +89,37 @@ public void HandleFailureMessage(string code, string message)
...
 
         public void HandleIgnoredMessage()
         {
-            _sentMessages.Dequeue();
-            _resultBuilders.Dequeue();
+            Unregister();
             _logger?.Debug("S: ", new IgnoredMessage());
         }
 
         public void Register(IRequestMessage requestMessage, IResultBuilder resultBuilder = null)
```
zhenlineo, line 102: Improve it to RegisterHandler or something

zhenlineo:
Move streaming into 1.1 first as it changes a lot https://github.com/neo4j/neo4j-dotnet-driver/pull/96

PR #97, opened 2016-07-13
buffer stream result

zhenlineo:
Buffer records within a session, but discard records if session is going to be returned to pool (RESET).
The current impl highly relies on the fact that in this driver `RUN` is synced immediately, while `PULL_ALL` is not. So we would have a result waiting to be filled if `PULL_ALL` has not been handled when we `RESET`. Interrupting this result receiving (setting this result to null) will result in no records being saved in memory. Right now, interrupt is performed only by `RESET`.

Note, if we run txs in a session, then the record will always be buffered if we commit the result. If we rollback the tx, then records get buffered if more txs afterwards, otherwise no record get buffered if we return the session to pool.

Session: 
RUN, sync, PULL_ALL, RESET, sync

Tx:
RUN_begin, DISCARD_ALL, RUN, sync, PULL_ALL, RUN_commit, DISCARD_ALL, **sync**, RESET, sync
RUN_begin, DISCARD_ALL, RUN, sync, PULL_ALL, RUN_rollback, DISCARD_ALL, RESET, sync

Where do we sync messages in this driver?
- INIT
- RUN (PULL_ALL)
- RUN commit, DISCARD_ALL
- RESET

What messages are enqueue but not synced?
- PULL_ALL
- ACK_FAIL
- RUN rollback, DISCARD_ALL
- DISCARD_ALL

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/ConnectionIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/ConnectionIT.cs
@@ -107,6 +107,55 @@ namespace Neo4j.Driver.IntegrationTests
         }
 
         [Fact]
+        public void ResultsHaveNotBeenReadGetLostAfterSessionClosed()
+        {
+            using (var driver = GraphDatabase.Driver(_serverEndPoint, _authToken))
+            {
+                IStatementResult result;
+                using (var session = driver.Session())
+                {
+                    result = session.Run("unwind range(1,3) as n RETURN n");
+                }
+                var resultAll = result.ToList();
+
+                // Records that has not been read inside session get lost
+                resultAll.Count.Should().Be(0);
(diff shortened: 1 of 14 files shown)


zhenlineo:
The feature `discard records if session is going to be returned to pool` is confusing actually, as it bring more trouble to understand when buffering and when not buffering. Also it bring questions like why tx.dispose will not throw away un-consumed records for previous txs. So I think I should change it to be in line with all other drivers that the buffering just kicks in when results are not fully consumed. And then add the smart `interrupt` back when we want to throw away records for both session.dispose and tx.dispose

zhenlineo:
Mind change again as when the server seen reset in a long running statement, it will stop sending more records to client anyway, so the result client received could actually be wrong (part of the real one). So I would say let's keep throw away the record for correctness.

PR #115, opened 2016-10-07
Routing driver

zhenlineo:
Created `ClusterConnectionPool` as pool of existing connection pools.

Added new scheme `bolt+routing` for creating a new `RoutingDriver` for routing queries to a cluster.
- The `RoutingDriver` holds a `Loadbalancer` which is responsible for supplying an usable connection during the creation of a specific read/write session.
- The `Loadbalancer` holds a `ClusterView` which describes the client's current understanding of the servers in the cluster and a `ClusterConnectionPool` that maintains a dictionary of {server, connection pool}.
- The servers in `ClusterView` might become unavailable and get (temporarily or permanently) removed. The `ClusterConnectionPool` could choose to either still keep the connection pool of a unavailable server or remove it totally too. However it is illegal to have servers that are known to `ClusterView` but have no connection pool entry in `ClusterConnectionPool`.
- The `ClusterView` load balance the connections to the servers using a `ConcurrentRoundRobinSet`.

Added error handling callback in `SocketConnection`.
- Moved the logic of verifying if a pooled connection is re-usable from `SocketConnection` to `PooledConnection`
- The purge of the servers and connection pools in `Loadbalancer` is done via a error handling event callback in `SocketConnection` passed by `LoadBalancer`.

TODO:
Feature: Retry rediscovery?
ITs


Neo4j.Driver/Neo4j.Driver/Internal/Routing/RoundRobinClusterView.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Neo4j.Driver.Internal.Routing
+{
+    internal class RoundRobinClusterView
+    {
+        private const int MinRouterCount = 1;
+        private readonly ConcurrentRoundRobinSet<Uri> _routers = new ConcurrentRoundRobinSet<Uri>();
+        private readonly ConcurrentRoundRobinSet<Uri> _detachedRouters = new ConcurrentRoundRobinSet<Uri>();
```
lutovich, line 28: Uris are only added to this set, otherwise it is never used. We need to decide if `_detachedRouters` is needed.

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/Neo4jErrorExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    internal static class Neo4jErrorExtensions
+    {
+        public static bool IsRecoverableError(this Neo4jException error)
+        {
+            return (error is ClientException || error is TransientException) && !IsClusterError(error);
+        }
+
+        public static bool IsClusterError(this Neo4jException error)
+        {
+            return error.Code.Equals("Neo.ClientError.Cluster.NotALeader")
```
lutovich: Would be nice to have constants or helper methods for these strings `Neo.ClientError.Cluster.NotALeader` and `Neo.ClientError.General.ForbiddenOnReadOnlyDatabase`. They are used here in `Neo4jErrorExtensions` and in `RoundRobinLoadBalancer`.

Neo4j.Driver/Neo4j.Driver/V1/GraphDatabase.cs
```diff
@@ -102,7 +103,22 @@ public static IDriver Driver(Uri uri, IAuthToken authToken, Config config = null
             config = config ?? Config.DefaultConfig;
             var encryptionManager = new EncryptionManager(config.EncryptionLevel, config.TrustStrategy, config.Logger);
             var connectionPoolSettings = new ConnectionPoolSettings(config.MaxIdleSessionPoolSize);
-            return new Internal.Driver(uri, authToken, encryptionManager, connectionPoolSettings, config.Logger);
+
+            if (uri.Port == -1)
+            {
+                var builder = new UriBuilder(uri.Scheme, uri.Host, 7687);
```
lutovich: Maybe let's introduce a constant for `7687`

lutovich (commented):
@zhenlineo changes look good to me

PR #118, opened 2016-11-08
Removing TOFU and TrustNonLocal

zhenlineo:


lutovich (commented):
@zhenlineo PR looks good. small question: shouldn't `TrustOnFirstUse` be deprecated first instead of being removed right away?

**Update:** TOFU is new for .Net driver and has never been released. thus it is ok to just remove it.

lutovich (approved):

PR #119, opened 2016-11-11
Add server info on result summary

zhenlineo:
Make the result to obtain the server info from the connection that it connects to.

PR #120, opened 2016-11-11
Bookmark

zhenlineo:
Based on https://github.com/neo4j/neo4j-dotnet-driver/pull/119


lutovich (changes requested):
@zhenlineo changes look fine but I think we need more tests. it seems `BookmarkCollector` class and `ISession.BeginTransaction()` method with non-null bookmark are not tested.

lutovich (approved):

PR #124, opened 2016-11-22
Always rollback tx when failed to commit

zhenlineo:
Fix the bug where if we forget to consume a tx in using with resources, then the tx is left without being rolled back if the tx failed to commit.

Also added back test `ShouldNotThrowExceptionWhenDisposeSessionAfterDriver` that is lost during refactoring

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriverIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriverIT.cs
@@ -298,11 +298,13 @@ namespace Neo4j.Driver.IntegrationTests
             }
 
             [Fact]
-            public void AfterErrorTheFirstSyncShouldAckFailureSoThatNewStatementCouldRunForTx()
+            public void RollBackTxIfErrorWithConsume()
             {
+                // Given
                 using (var driver = GraphDatabase.Driver(_serverEndPoint, _authToken))
                 using (var session = driver.Session())
                 {
+                    // When failed to run a tx with consume
                     using (var tx = session.BeginTransaction())
                     {
                         var ex = Record.Exception(() => tx.Run("Invalid Cypher").Consume());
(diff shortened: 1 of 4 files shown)


Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriverIT.cs
```diff
@@ -298,41 +298,75 @@ public void AfterErrorTheFirstSyncShouldAckFailureSoThatNewStatementCouldRun()
...
+                    // The following code is the same as using(var tx = session.BeginTx()) {...}
+                    // While we have the full control of where the error is thrown
+                    var tx = session.BeginTransaction();
+                    tx.Run("Invalid Cypher");
+                    tx.Success();
+                    var ex = Record.Exception(() => tx.Dispose());
+                    ex.Should().BeOfType<ClientException>();
+                    ex.Message.Should().StartWith("Invalid input 'I'");
+
+                    // Then can still run more afterwards
```
technige, line 340: To check whether rollback has occurred more directly, the transaction could contain a `CREATE` followed by invalid Cypher. The test would then assert that the creation had been undone.

Neo4j.Driver/Neo4j.Driver/Internal/Transaction.cs
```diff
@@ -90,27 +90,47 @@ protected override void Dispose(bool isDisposing)
...
             {
-                _cleanupAction.Invoke();;
+                _cleanupAction.Invoke();
                 base.Dispose(true);
             }
         }
 
+        private void RollBackTx()
+        {
+            _connection.Run(Rollback, null, new BookmarkCollector(s => Bookmark = s));
```
technige: `ROLLBACK` will never produce a bookmark.

zhenlineo: Oh, that makes sense. Similarly here, we should do `Collector.NO_OP`
 https://github.com/neo4j/neo4j-java-driver/blob/1.1/driver/src/main/java/org/neo4j/driver/internal/ExplicitTransaction.java#L161

PR #131, opened 2016-12-20
Changed the examples to use using for driver and session resources

zhenlineo:
Removed using driver if not needed
Changed driver to be a local var


lutovich:
@zhenlineo looks good to me. One question: `_driver` is now never disposed. Is this fine?

PR #134, opened 2016-12-23
Added connection timeout

zhenlineo:
Default to 5s
Abstracted ConnectionSettings class to include all related settings for a connection inside and pass through to SocketConnection

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionSettings.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionSettings.cs
@@ -1,0 +1,46 @@
+﻿// Copyright (c) 2002-2016 "Neo Technology,"
+// Network Engine for Objects in Lund AB [http://neotechnology.com]
+// 
+// This file is part of Neo4j.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Neo4j.Driver.V1;
+
+namespace Neo4j.Driver.Internal
+{
+    internal class ConnectionSettings
+    {
+        internal const string DefaultUserAgent = "neo4j-dotnet/1.1";
+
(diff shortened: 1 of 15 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionSettings.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public Uri InitialServerUri { get; }
+        public IAuthToken AuthToken { get; }
+        public TimeSpan ConnectionTimeout { get; }
+        public string UserAgent { get; }
+        public EncryptionManager EncryptionManager;
+
+        public ConnectionSettings(Uri initialServerUri, IAuthToken authToken, EncryptionManager encryptionManager, TimeSpan connectionTimeout, string userAgent = null)
+        {
+            Throw.ArgumentNullException.IfNull(initialServerUri, nameof(initialServerUri));
```
lutovich, line 35: Could `Contract.Requires` be used here? As described here https://msdn.microsoft.com/en-us/library/dd264808(v=vs.110).aspx

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -58,7 +71,11 @@ public void Init()
         {
             try
             {
-                Task.Run(() => _client.Start()).Wait();
+                var connected = Task.Run(() => _client.Start()).Wait(_connectionTimeout);
```
lutovich, line 74: Shouldn't we cancel the `_client.Start()` task if `!connected`?
Could it still remain running in the background?
I do not know for sure, reading these articles:
https://msdn.microsoft.com/en-us/library/mt674886.aspx
https://msdn.microsoft.com/en-us/library/system.threading.cancellationtokensource(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/dd235606(v=vs.110).aspx

Would be nice to have this tested as well.

lutovich (changes requested):
@zhenlineo review round completed

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionSettings.cs
zhenlineo: I did not find this lib in .net core libs

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
zhenlineo: We will close the whole connection (where we call `_client.Dispose()`) in `ConnectionPool.Acquire` if failed to create a new connection.

lutovich (approved):

PR #135, opened 2017-01-05
Improved tests for bookmark

zhenlineo:
Added tests to check that bookmark is set to null when explicit transaction is rolled back or fails.


zhenlineo:
closing this one as it already included in another PR

PR #133, opened 2016-12-22
Added ITs for routing driver

zhenlineo:
Added basic ITs for testing routing driver
Added boltkit support to start/stop a local cluster

PR #139, opened 2017-01-25
Fix the bug in load balancer that the new connection pools are not connecting to the specified server

zhenlineo:
Fix the bug where the connection pool was always using the initial uri rather than the new uri specified by load balancer when creating a new connection pool

Removed some compilation warnings


zhenlineo:
The test for this fix is in #133

lutovich (approved):

PR #142, opened 2017-02-03
Change the routing according to the specification

zhenlineo:


Neo4j.Driver/Neo4j.Driver.Tests/Routing/LoadBalancerTests.cs
```diff
@@ -0,0 +1,561 @@
...
+                {
+                    // Given
+                    var routingTable = NewRoutingTable();
+                    var balancer = new LoadBalancer(null, routingTable);
+
+                    // When
+                    var error = Record.Exception(()=>AcquiredConn(balancer, mode));
+
+                    // Then
+                    error.Should().BeOfType<SessionExpiredException>();
```
lutovich, line 424: Shouldn't we throw `ServiceUnavailableException` here?

Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
```diff
@@ -55,12 +56,20 @@ private Stream Stream
...
+                            (sender, certificate, chain, errors) =>
+                                _encryptionManager.TrustStrategy.ValidateServerCertificate(_uri, certificate, errors));
+                        var sslTask = secureStream.AuthenticateAsClientAsync(_uri.Host, null,
+                            System.Security.Authentication.SslProtocols.Tls12, false);
+                        sslTask.Wait();
+                        _stream = secureStream;
+                    }
+                    catch (Exception e)
+                    {
+                        throw new SecurityException($"Failed to establish encrypted connection with server {_uri}.", e);
```
lutovich: Do we maybe need to try to close `_stream` if exception happens?

lutovich (commented):
@zhenlineo looks good. Made two small comments/questions

Neo4j.Driver/Neo4j.Driver.Tests/Routing/LoadBalancerTests.cs
zhenlineo: This class is to test the acquireReadConn or acquireWriteConn methods without ensureRouting. That is to say if we already assume that the routing table we obtained is a valid one, then test what conn we will obtain from acquireRead/WriteConn method call. 

This method tests that if the routing table is valid but we failed to connect to any of the server in routing table, then we should session expired and let user retry.

If we consider ensureRouting, then this should definitely be a SU.

PR #143, opened 2017-02-07
Better error handling

zhenlineo:
Based on #142 

* Made `DirectDriver` throw `ServiceUnavailable` when failed to send/receive data caused by `IOException` and `SocketException`.
* Ensured connections who have connection errors always closed immediately when the error is received. Also ensure connections where `ProtocolError` is received from server will be immediately stop too. 
* Ensured connections who have connection errors always mark connections not recoverable. (A fallback to ensure the connection will be closed again when session.close)
* Changed most `ArgumentOutOfRangeException` in `PackStream` to `ProtocolException`.

* Improved the IT structure to only start corresponding neo4j services for stand alone or cluster.
* Changed the error handling logic on connections. Before each connection could register multiple error handlers, while no order among the handlers is guaranteed. Now each connection could accept only one external error hander. The external handler will only be executed after internal handler has been called.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Extensions/Throw.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Extensions/Throw.cs
@@ -36,7 +36,7 @@ namespace Neo4j.Driver.Internal
             }
         }
 
-        public static class ArgumentException
+        public static class ProtocolException
         {
             public static void IfNotEqual(int first, int second, string firstParam, string secondParam)
             {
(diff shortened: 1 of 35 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Extensions/Throw.cs
```diff
@@ -54,13 +54,13 @@ internal static void IfNotEqual(object first, object second, string firstParam,
             public static void If(Func<bool> func, object first, object second, string firstParam, string secondParam)
             {
                 if(func())
-                    throw new System.ArgumentException($"{firstParam} ({first}) did not equal {secondParam} ({second})");
+                    throw new V1.ProtocolException($"{firstParam} ({first}) does not equal to {secondParam} ({second})");
             }
 
             public static void IfNotTrue(bool value, string nameofValue)
```
lutovich: I'd vote for `IfFalse` to remove negation

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/Throw.cs
```diff
@@ -77,6 +77,11 @@ public static void IfValueGreaterThan(long value, long limit, string parameterNa
                 if(value > limit)
                     throw new System.ArgumentOutOfRangeException(parameterName, value, $"Value given ({value}) cannot be greater than {limit}.");
             }
+            public static void IfNotTrue(bool value, string nameofValue)
```
lutovich: Maybe also `IfFalse`

lutovich (approved):

Issue #141, opened 2017-02-02
Excessive Open Connections


roadbikemike:
We are having some major issues with the .NET Neo driver (v1.1).  When running tests from a console app, the connections are managed as expected.  However, the same code on IIS and exercised via JMeter is completely different.  When the code hits an exception, in our case "Neo4j.Driver.V1.ClientException: Failed to connect to the server [some ip]:7687 within connection timeout 5000ms" the connection is not closed and they keep on rising and rising up in to the 1000's.

I have tried setting MaxIdleSessionPoolSize anywhere between 10 and 100 with no difference.

Here is my code:

string username = ConfigurationManager.AppSettings["NeoUser"]; ;
string password = ConfigurationManager.AppSettings["NeoPass"]; ;
string boltUri = ConfigurationManager.AppSettings["boltUri"];
string PooledConns = ConfigurationManager.AppSettings["PooledCons"];

var config = new Config { };
config.MaxIdleSessionPoolSize = int.Parse(PooledConns);

	try
	{
		using (IDriver driver = GraphDatabase.Driver(boltUri, AuthTokens.Basic(username, password), config))
		{
			using (var session = driver.Session())
			{

				var setGUID = Guid.NewGuid().ToString();

				IDictionary<string, object> param = new Dictionary<string, object>();
				param.Add("0", "User-" + setGUID);
				param.Add("1", "Display-" + setGUID);
				param.Add("2", "Name=" + setGUID);

				session.Run("CREATE(a: User { id: {0}, displayName: {1}, identity: {2}})", param).Consume();

				return Request.CreateResponse(HttpStatusCode.OK, "User successfully added!");

			}
		}
	}
	catch (Exception ex)
	{
		Logger.Write("---------------------------------------");
		Logger.Write(ex.ToString());

		return Request.CreateResponse(HttpStatusCode.InternalServerError, ex.ToString());
	}
}

zhenlineo:
Hi, I think the problem that you describe indicates that we did not close the connection when failed to connect. A quick look into the code indicates that we actually will aways close the connection on failed connection initialization before re-throw the error. 

Setting the maxIdleSessionPoolSize will not help too much on this problem as this config will not affect how many connections could be created but how many could be buffered.

I will look into you problem and give you a fix asap.

Zhen

roadbikemike:
[...]

PR #144, opened 2017-02-07
Connection leak investigation

zhenlineo:
Investigating on connection leak in #141

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
@@ -19,67 +19,51 @@ using System.IO;
 using System.Net.Security;
 using System.Net.Sockets;
 using System.Threading.Tasks;
+using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal.Connector
 {
     internal class TcpSocketClient : ITcpSocketClient
     {
         private readonly TcpClient _client;
-        private bool _useTls;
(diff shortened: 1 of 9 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
```diff
@@ -79,7 +79,7 @@ private Stream Stream
         {
             _uri = uri;
             _useTls = useTlsEncryption;
-            await _client.ConnectAsync(uri.Host, uri.Port).ConfigureAwait(true);
+            await _client.ConnectAsync(uri.Host, uri.Port).ConfigureAwait(false);
```
zhenlineo: @cskardon Hello, I am looking at this code that you added before, but I do not understand why this await has to go back to its calling context. Do you still remember why configAwait has to be `true` here? Can I change it to `false`?

cskardon: Errrrr - if I'm totally honest with you I have no idea why that is set to `true` - I would have thought we'd have put it to `false`. There's no reason why we'd need to stay in the same SyncContext - I don't even know why we'd have done that in the first place. 

Does it crop up anywhere else in the codebase?

zhenlineo: No, this is the only place to have this value to be true. So I guess I will just change it to false then.

PR #146, opened 2017-02-16
Enable socket keep alive option

zhenlineo:


Neo4j.Driver/Neo4j.Driver/V1/Config.cs
```diff
@@ -112,6 +113,11 @@ static Config()
         /// </summary>
         public TimeSpan ConnectionTimeout { get; set; } = TimeSpan.FromSeconds(5);
 
+        /// <summary>
+        /// Gets or sets the socket keep alive option.
+        /// </summary>
+        public bool SocketKeepAlive { get; set; } = true;
```
lutovich, line 119: Should we maybe expose this option with name `ConnectionKeepAlive`?

lutovich (commented):
@zhenlineo looks good, one small naming question

PR #147, opened 2017-02-17
Refactoring on connection pool

zhenlineo:
Made conneciton pool lock free
Added ConcurrentSet impl based on ConcurrentDict
Improved the impl of RRSet to use list to avoid memory copy

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConcurrentSet.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConcurrentSet.cs
@@ -1,0 +1,49 @@
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Neo4j.Driver.Internal
+{
+    internal class ConcurrentSet <T> : IEnumerable<T>
+    {
+        private readonly ConcurrentDictionary<T, bool> _dictionary;
+
+        public ConcurrentSet()
+        {
+            _dictionary = new ConcurrentDictionary<T, bool>();
+        }
+
+        /// <summary>
+        /// true if the item was added to the set successfully; false if the item already exists.
+        /// </summary>
+        /// <param name="item"></param>
+        /// <returns></returns>
+        public bool TryAdd(T item)
+        {
+            return _dictionary.TryAdd(item, true);
+        }
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConcurrentSet.cs
```diff
@@ -0,0 +1,49 @@
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Neo4j.Driver.Internal
+{
+    internal class ConcurrentSet <T> : IEnumerable<T>
```
lutovich, line 7: Should `ConcurrentSet` maybe implement `ISet`?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -223,24 +217,20 @@ protected override void Dispose(bool isDisposing)
...
+                foreach (var inUseConnection in _inUseConnections)
                 {
-                    var connections = new List<IPooledConnection>(_inUseConnections.Values);
-                    _inUseConnections.Clear();
-                    foreach (var inUseConnection in connections)
+                    Logger?.Info($"Disposing In Use Connection {inUseConnection.Id}");
+                    if (_inUseConnections.TryRemove(inUseConnection))
                     {
-                        Logger?.Info($"Disposing In Use Connection {inUseConnection.Id}");
                         inUseConnection.Close();
```
lutovich, line 225: Can this call throw an exception?

lutovich (approved):
@zhenlineo changes look good. Made couple small questions.

Neo4j.Driver/Neo4j.Driver/Internal/ConcurrentSet.cs
zhenlineo: We could, then we need to impl 20+ methods on ISet interface. I feel it is not needed.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: In theory, no, as all impl of `IConnection.Close` method will only log errors but not throw any out.

PR #148, opened 2017-02-20
Added `ClusterConnection` to abstract error handling out of `SocketConnection`

zhenlineo:
Based on #147 
Before, the error handling logic for removing the pooled connection from the pool and the logic for purging from load balancer was all added on `SocketConnection`. This PR moved error handling logic to the Connection that it belongs to.


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -85,12 +80,9 @@ private IPooledConnection CreateNewPooledConnection()
             try
             {
                 conn = _fakeConnection != null
-                    ? new PooledConnection(_fakeConnection, Release)
-                    : new PooledConnection(new SocketConnection(_uri, _connectionSettings, _logger), Release);
-                if (_externalErrorHandler != null)
-                {
-                    conn.ExternalConnectionErrorHander(_externalErrorHandler);
-                }
+                    ? new PooledConnection(()=>_fakeConnection, Release)
```
lutovich, line 83: Is it possible to create overridable `createConnection()` method instead of having this `_fakeConnection` field?

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/ErrorExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public static bool IsClusterError(this Exception error)
+        {
+            return IsClusterNotALeaderError(error)
+                   || IsForbiddenOnReadOnlyDatabaseError(error);
+        }
+
+        public static bool IsClusterNotALeaderError(this Exception error)
+        {
+            var exception = error as Neo4jException;
```
lutovich: Maybe extract a `hasErrorCode(Exception, string)` method and call it from `IsClusterNotALeaderError` and `IsForbiddenOnReadOnlyDatabaseError`

lutovich (approved):
@zhenlineo looks good, made two small comments.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Good idea. I think the initial idea was to avoid using `virtual` modifier. But overall it should make this class easier. I will make this change outside this PR.

PR #149, opened 2017-02-21
Decouple conn from session

zhenlineo:
based on #147 

* `Connection` used by `session.Run` and/or `transaction` should be disposed one and only once.
* `Connection` used by `session.Run`:
  - It will be disposed when the reply message (`SUCCESS/IGNORE/FAILURE`) of `PULL_ALL` is received.
  - It will also be disposed in `session.Dispose` and/or another new `session.Run` and/or a new `transaction` if no reply message of `PULL_ALL` received.

* `Connection` used by transaction will be disposed when the tx is disposed or the session is disposed.

* Ensure a connection cannot be created after `driver.Dispose`

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
@@ -21,8 +21,9 @@ namespace Neo4j.Driver.IntegrationTests
         {
             Exception exception;
             using (var driver = GraphDatabase.Driver("bolt://localhost:123"))
+            using (var session = driver.Session())
             {
-               exception = Record.Exception(()=>driver.Session());
+                exception = Record.Exception(() => session.Run("RETURN 1"));
             }
             exception.Should().BeOfType<ServiceUnavailableException>();
             exception.Message.Should().Be("Connection with the server breaks due to AggregateException: One or more errors occurred.");
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
@@ -57,7 +57,7 @@ namespace Neo4j.Driver.IntegrationTests
 
             var error = Record.Exception(() => session.Run("RETURN 1"));
             error.Should().BeOfType<ObjectDisposedException>();
-            error.Message.Should().StartWith("Cannot acquire a new connection as ConnectionPool has already been disposed.");
+            error.Message.Should().StartWith("Failed to acquire a new connection as the driver has already been disposed.");
         }
 
         [Fact]
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
@@ -47,6 +47,20 @@ namespace Neo4j.Driver.IntegrationTests
         }
 
         [Fact]
+        public void DisallowRunInSessionAfterDriverDispose()
+        {
+            var driver = GraphDatabase.Driver(ServerEndPoint, AuthToken);
+            var session = driver.Session(AccessMode.Write);
+            session.Run("RETURN 1").Single()[0].ValueAs<int>().Should().Be(1);
+
+            driver.Dispose();
+
+            var error = Record.Exception(() => session.Run("RETURN 1"));
+            error.Should().BeOfType<ObjectDisposedException>();
+            error.Message.Should().StartWith("Cannot acquire a new connection as ConnectionPool has already been disposed.");
+        }
+
+        [Fact]
         public void ShouldConnectAndRun()
         {
             using (var session = Driver.Session())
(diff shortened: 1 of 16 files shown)


Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
```diff
@@ -46,6 +47,20 @@ public void DisallowNewSessionAfterDriverDispose()
...
+        public void DisallowRunInSessionAfterDriverDispose()
+        {
+            var driver = GraphDatabase.Driver(ServerEndPoint, AuthToken);
+            var session = driver.Session(AccessMode.Write);
+            session.Run("RETURN 1").Single()[0].ValueAs<int>().Should().Be(1);
+
+            driver.Dispose();
+
+            var error = Record.Exception(() => session.Run("RETURN 1"));
+            error.Should().BeOfType<ObjectDisposedException>();
```
lutovich, line 59: Is it possible to throw some sort of `DriverDisposedException` in this case? Is it the only way connection pool can be disposed?

Neo4j.Driver/Neo4j.Driver.IntegrationTests/RoutingDriver/RoutingDriverIT.cs
```diff
@@ -124,8 +128,8 @@ public void ShouldDisallowMoreStatementAfterDriverDispose()
 
             driver.Dispose();
             var error = Record.Exception(() => session.Run("RETURN 1"));
-            error.Should().BeOfType<ClientException>();
-            error.Message.Should().Contain("The current session cannot be reused as the underlying connection with the server has been closed");
+            error.Should().BeOfType<ObjectDisposedException>();
+            error.Message.Should().StartWith("Cannot acquire a new connection as LoadBalancer has already been disposed.");
```
lutovich: Would be nice to signal with this message that driver is disposed. "LoadBalancer" could be confusing for users.

Neo4j.Driver/Neo4j.Driver.Tests/SessionTests.cs
```diff
@@ -95,33 +95,34 @@ public void ShouldBeAbleToUseSessionAgainWhenTransactionIsClosed()
...
             {
                 var mockConn = new Mock<IConnection>();
-                mockConn.Setup(x => x.IsOpen).Returns(false);
-                var session = new Session(mockConn.Object, null);
+                var session = new Session(mockConn.Object);
+                session.Run("lalal");
 
-                var error = Record.Exception(() => session.Run("lalal"));
-                error.Should().BeOfType<ClientException>();
+                session.Run("bibib");
```
lutovich, line 112: I liked "lalal" :(

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -51,24 +52,42 @@ internal class LoadBalancer : ILoadBalancer
...
             }
+
+            if (_disposeCalled)
+            {
+                ThrowObjectClosedException();
+            }
+            return conn;
+        }
+
+        private void ThrowObjectClosedException()
```
lutovich: Maybe `ThrowObjectDisposedException`

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -68,97 +98,104 @@ protected override void Dispose(bool isDisposing)
...
-        private void EnsureSessionIsOpen()
+        /// <summary>
+        /// Clean any session.run result reference.
+        /// If session.run result is not fully consumed, then pull full result into memory.
+        /// </summary>
+        /// <exception cref="ClientException">If error when pulling result into memory</exception>
+        private void DisposeSessionResult()
         {
-            if (!_isOpen)
+            if (_sessionRunResult != null)
```
lutovich, line 162: Can we instead just check `_connection != null` and remove `_sessionRunResult` field?

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -68,97 +98,104 @@ protected override void Dispose(bool isDisposing)
...
-                                          "Make sure that you do not have a bad reference to a disposed session " +
-                                          "and retry your statement in another new session.");
+                try
+                {
+                    // this will enfore to buffer all unconsumed result
+                    _connection.Sync();
+                }
+                catch (Exception e)
+                {
+                    throw new ClientException($"Error when pulling unconsumed session.run records into memory in session: {e.Message}", e);
```
lutovich, line 171: Can we just propagate `e` and not wrap it in `ClientException`?

lutovich (commented):
@zhenlineo review round completed

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
zhenlineo: Would it be strange that you get the error of previous run in a new run? This wrapper is for not confusing a user where the error comes from and also try to educate the user that he should consume his result.

Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/SessionIT.cs
zhenlineo: Let's use this `ObjectDisposedException` for now and see if we want a new one latter.

Neo4j.Driver/Neo4j.Driver.Tests/SessionTests.cs
zhenlineo: Will try meow and woow next time

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
zhenlineo: Yes, but I keep a reference to result just to keep track that whose connection I am disposing now. Would like to attach some debugging tool later for better debugging.

lutovich (approved):

PR #150, opened 2017-02-22
Use init uri after service unavailable

zhenlineo:
[Ref](https://msdn.microsoft.com/en-us/library/ms143998(v=vs.110).aspx)

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Routing/IRoutingTable.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Routing/IRoutingTable.cs
@@ -12,6 +12,7 @@ namespace Neo4j.Driver.Internal.Routing
         void Remove(Uri uri);
         ISet<Uri> All();
         void Clear();
-        void EnsureRouter(IEnumerable<Uri> ips);
+        void AddRouter(IEnumerable<Uri> ips);
+        bool HasNoRouter();
     }
 }
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Routing/IRoutingTable.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Routing/IRoutingTable.cs
@@ -12,5 +12,6 @@ namespace Neo4j.Driver.Internal.Routing
         void Remove(Uri uri);
         ISet<Uri> All();
         void Clear();
+        void EnsureRouter(IEnumerable<Uri> ips);
     }
 }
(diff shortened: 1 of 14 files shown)


lutovich:
@zhenlineo needs rebase after merging #149

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -190,6 +197,13 @@ internal IRoutingTable UpdateRoutingTable(Func<IConnection, IRoutingTable> redis
             }
         }
 
+        private void EnsureInitialRouter()
+        {
+            var ips = _seed.ToIps();
+            _routingTable.EnsureRouter(ips);
+            _clusterConnectionPool.Add(ips);
```
lutovich: Can we use `_clusterConnectionPool.Update(ips)` instead of adding new `#Add()` method?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -190,6 +197,13 @@ internal IRoutingTable UpdateRoutingTable(Func<IConnection, IRoutingTable> redis
             }
         }
 
+        private void EnsureInitialRouter()
+        {
+            var ips = _seed.ToIps();
```
lutovich: Maybe null check for `_seed`? It's default value is null.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/RoundRobinRoutingTable.cs
```diff
@@ -104,5 +102,13 @@ public override string ToString()
                    $"[{nameof(_readers)}: {_readers}], " +
                    $"[{nameof(_writers)}: {_writers}]";
         }
+
+        public void EnsureRouter(IEnumerable<Uri> ips)
+        {
+            if (_routers.Count == 0)
```
lutovich: This method should be called only when `_routers.Count == 0` right? Maybe throw exception in other case?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/IRoutingTable.cs
```diff
@@ -12,5 +12,6 @@ internal interface IRoutingTable
         void Remove(Uri uri);
         ISet<Uri> All();
         void Clear();
+        void EnsureRouter(IEnumerable<Uri> ips);
```
lutovich: Can this method be called `AddRouters`? I do not understand the "ensure" part

lutovich (commented):
@zhenlineo review completed, couple comments

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
zhenlineo: The seed will be init by `ConnectionSettings.InitialServerUri`, where it will ensure no null value is provided. Though there is another constructor for testing that bypass init of `_seed`.

PR #151, opened 2017-02-28
Changed from passing methods to registered handlers

zhenlineo:
Fixed a few bugs:
* If a connection is broken in Session, then session.close will dispose the connection directly without sync.
* When tx receives an error when running a statement, then the tx will be marked to not be able to run more statements but wait to be disposed directly.
* Fixed the error check depending on the connection is a READ connection or WRITE connection in ClusterConnection used by Loadbalancer.


Neo4j.Driver/Neo4j.Driver/Internal/Extensions/CollectionExtensions.cs
```diff
@@ -15,12 +15,49 @@
...
 using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
+using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal
 {
     internal static class CollectionExtensions
     {
+        public static T[] DequeueToArray<T>(this Queue<T> queue, int length)
```
lutovich, line 26: `Queue` has `#ToArray()` method. Is this one different?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -229,37 +202,24 @@ private void EnsureInitialRouter()
...
-            }
-            throw e;
+            _logger?.Info($"Server at {uri} is no longer available due to error: {e.Message}.");
+            _routingTable.Remove(uri);
+            _clusterConnectionPool.Purge(uri);
+        }
+
+        public void OnWriteError(Uri uri)
+        {
+            _routingTable.Remove(uri);
```
lutovich: Shouldn't we only remove writer here?

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -176,27 +180,26 @@ private void DisposeSessionResult()
...
-                try
-                {
-                    _transaction.Dispose();
-                }
-                catch (Exception e)
-                {
-                    throw new ClientException($"Error when disposing unclosed transaction in session: {e.Message}", e);
-                }
-            }
+            LastBookmark = null;
```
lutovich, line 191: Maybe we should not set last bookmark to null here?

lutovich (commented):
@zhenlineo review completed, couple comments

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/CollectionExtensions.cs
zhenlineo: This one will remove the item from the queue too. Queue.ToArray will just get a copy without removing any

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
zhenlineo: This is changed in next PR for bookmark.

PR #152, opened 2017-02-28
causal consistency within a session

zhenlineo:
based on #151


Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/BookmarkIT.cs
```diff
@@ -96,7 +97,7 @@ public void ShouldThrowForInvalidBookmark()
         {
             if (Skip) return;
             var invalidBookmark = "invalid bookmark format";
-            using (var session = Driver.Session())
+            using (var session = (Session)Driver.Session())
```
lutovich, line 100: Why are these casts needed?

lutovich (approved):
@zhenlineo looks good, one small question

Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/BookmarkIT.cs
zhenlineo: To avoid compiling warnings :) The methods are not deprecated on Session class, they are deprecated on ISession

PR #161, opened 2017-04-05
Made beginTx(bookmark) ignoring null bookmark

zhenlineo:
Added more tests to make sure session always closes connection when failing creating transaction in beginTx

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
@@ -50,7 +50,7 @@ namespace Neo4j.Driver.Internal
             _logger = logger;
 
             _defaultMode = defaultMode;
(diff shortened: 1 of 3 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -195,6 +191,18 @@ public void OnTransactionDispose()
         }
 
         /// <summary>
+        /// Only set the bookmark to a new value if the new value is not null
+        /// </summary>
+        /// <param name="bookmark">The new bookmark</param>
+        private void Bookmark(string bookmark)
```
lutovich: Maybe `UpdateBookmark` or `SetBookmark`?

lutovich (approved):
@zhenlineo looks good, one small question.

PR #162, opened 2017-04-05
Do not retry on TransientError caused by user terminating tx

zhenlineo:


Neo4j.Driver/Neo4j.Driver/Internal/RetryLogic.cs
```diff
@@ -66,11 +69,12 @@ public T Retry<T>(Func<T> runTxFunc)
                 {
                     return runTxFunc();
                 }
-                catch (Exception e) when (e is SessionExpiredException || e is TransientException || e is ServiceUnavailableException)
+                catch (Exception e) when (e is SessionExpiredException || e.IsRecoverableError() || e is ServiceUnavailableException)
```
lutovich: Should newly introduced `#IsRetriableTransientError()` be used here instead of `#IsRecoverableError()`?
Maybe also add tests asserting that retries do not happen on `Terminated` and `LockClientStopped`.

lutovich (changes requested):
@zhenlineo review completed.

zhenlineo:
@lutovich Test added to address the wrong method call

lutovich (approved):

Issue #145, opened 2017-02-14
Socket Errors after Idle


roadbikemike:
When we are running load tests against Neo using the .NET 1.1.1 driver, there are no issues.  We can push thousands of requests through as expected.  However, if we let our application sit idle and then try a connect, we get the following below.  In this case, I let Neo and the client application sit over night with no activity.  Yesterday, before leaving everything was working great, but throwing these errors this morning.  After a restart of the client side application, things were again working as intended.  It seems as if the driver is leaving something open and expecting the socket to be alive when it is not.

**Client Side**:
2017-02-14 13:17:08,584 [18] ERROR NeoDrivergFileAppender - Unable to unpack message from server, connection has been terminated.
System.IO.IOException: Unable to read data from the transport connection: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. ---> System.Net.Sockets.SocketException: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond
   at System.Net.Sockets.Socket.Receive(Byte[] buffer, Int32 offset, Int32 size, SocketFlags socketFlags)
   at System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size)
   --- End of inner exception stack trace ---
   at System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size)
   at Neo4j.Driver.Internal.Connector.SocketExtensions.Read(Stream stream, Byte[] bytes)
   at Neo4j.Driver.Internal.Connector.ChunkedInputStream.ReadSpecifiedSize(Byte[] buffer)
   at Neo4j.Driver.Internal.Connector.ChunkedInputStream.Ensure(Int32 size)
   at Neo4j.Driver.Internal.Connector.ChunkedInputStream.ReadByte()
   at Neo4j.Driver.Internal.Packstream.PackStream.Unpacker.UnpackStructHeader()
   at Neo4j.Driver.Internal.Packstream.PackStreamMessageFormatV1.ReaderV1.Read(IMessageResponseHandler responseHandler)
   at Neo4j.Driver.Internal.Connector.SocketClient.ReceiveOne(IMessageResponseHandler responseHandler)

**Server Side**:
[...]

PR #165, opened 2017-04-10
Idle timer on each pooled connection to record the idle time elapsed

zhenlineo:
By default the connection pool will not timeout any connection inside the pool. As long as the connection is still "open", then the connection is reused.
While a user could enable idle timeout on connections by setting `ConnectionIdleTimeout` to be a non-negative number.
Then when reusing connections from the pool, connections that have been idle for longer than the configured time will be closed directly.

Notes:
* The timer is implemented using `Stopwatch`, the time elapsed is loosely precise.
* The connection will not immediately be closed when idle timeout reached, instead it will be closed in the same thread that get it out of the pool.


zhenlineo:
@cskardon I am trying to start solving #145 by making smaller changes. So this PR added a simple timer on each connection and let the pool close the connections that has been idle for too long in the caller thread.

It would be great if we do not need to introduce another thread to close connections.

Issue #164, opened 2017-04-10
Asp dot net core bolt driver can't connect on *nix


ChaosYoda:
When trying to run an app on *nix with dot net core, I get This platform does not support connecting sockets to DNS endpoints via the instance Connect and ConnectAsync methods, due to the potential for a host name to map to multiple IP addresses and sockets becoming invalid for use after a failed connect attempt. Use the static ConnectAsync method, or provide to the instance methods the specific IPAddress desired. This is related to https://github.com/dotnet/corefx/issues/8768 and https://github.com/StackExchange/StackExchange.Redis/issues/463. There is a work around but currently the issue seems to be in the driver.

zhenlineo:
Thanks for reporting this issue. I was recently looking into resolving DNS to IPs before creating connection pool for this driver. I think your problem will be solved by introducing Dns resolving. I was holding doing it as I was a bit reluctant to introduce another async dns.ResolveAsync into this synchronized driver. 

While waiting for a fix, I would also suggest some work around for now. Consider to provide IP address directly to the driver rather than a Dns address. If you want to use multi IPs to avoid single point failure, then consider to put a load-balancer in the middle and pass the IP of load-balancer to the driver.

ChaosYoda:
Hi, I am using the IP directly and it gives me the error above.

zhenlineo:
Oh I see, so there is a problem with 
```
public Task ConnectAsync(string host, int port);
```

 and we should avoid the problem by moving to 

```
public Task ConnectAsync(IPAddress address, int port);
public Task ConnectAsync(IPAddress[] addresses, int port);
```

Am I got you correctly this time?

ChaosYoda:
Yes I think that will work.

zhenlineo:
Hi, the problem should go away with the new .net [1.3.0-beta01](https://github.com/neo4j/neo4j-dotnet-driver/releases/tag/1.3.0-beta01). The fix will also be included in 1.2.1 (which should be coming soon).

PR #167, opened 2017-04-11
resolve dns to ip addresses when establishing socket connection

zhenlineo:
When establish a socket connection, instead of using `(Uri.Host, Uri.Port)`, we will use `(IPAddress, port)` instead to avoid issues on some platform where socket cannot be established with dns addresses due to one to many mapping.

If the uri given is already an ip address, then the ip address will be used directly. Otherwise, we will perform a dns lookup to resolve to ip addresses. The resolved ip addresses will be put in an array where IPv6 addresses will be placed first. Then socket client will try to establish connections with the server's ip addresses in order until connection established successfully.

This change does not implement connection pool level dns resolution. The connection pool still holds a pool for each uri (could either be a dns or ip address). While this PR does not need to be modified when connection pool level dns lookup is added as the socket client will not do a dns resolve if the uri is already an ip address.

Fix for #164 
Based on #166


Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
```diff
@@ -73,6 +72,26 @@ public TcpSocketClient(EncryptionManager encryptionManager, bool keepAlive, ILog
             }
         }
 
+        private async Task Connect(Uri uri)
+        {
+            var addresses = await uri.ResolveAsyc();
```
lutovich, line 77: We should probably log these addresses in debug mode and include them in the `IOException`

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/NetworkExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            else
+            {
+                addresses = (await Dns.GetHostAddressesAsync(uri.Host).ConfigureAwait(false))
+                    .OrderBy(x=>x, new AddressComparer(AddressFamily.InterNetworkV6)).ToArray();
+            }
+            return addresses;
+        }
+
+        private class AddressComparer : IComparer<IPAddress>
```
lutovich: Could you please add some unit tests for `AddressComparer`?

lutovich (approved):
@zhenlineo looks good, two small comments.

PR #168, opened 2017-04-12
Routing context support in uri

zhenlineo:
Made driver aware getRoutingTable in 3.2+ server
Made driver parse routing context from uri


Neo4j.Driver/Neo4j.Driver.Tests/TestUtil/NetworkExtensionsTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using System;
+using FluentAssertions;
+using Neo4j.Driver.Internal;
+using Xunit;
+
+namespace Neo4j.Driver.Tests
+{
+    public class NetworkExtensionsTests
+    {
+        public class ParseRoutingContextMethod
```
lutovich: Maybe also include tests without port, without routing params and for uri like this `bolt://localhost/?` because it is also a valid one.

lutovich (approved):
@zhenlineo looks good, one comment about additional tests.

PR #170, opened 2017-04-18
Allow the driver to serve read operations when the writer is absent.

zhenlineo:
In order to do this, a few driver behaviours have been changed:
* Move the EnsureRoutingTable check inside each AcquireReader/Writer method
* Allow the driver to accept routing table wihtout a writer
* Always first try initial routers for a new routing table in next discovery call if the writer is absent


Neo4j.Driver/Neo4j.Driver.Tests/Routing/ConcurrentRoundRobinSetTests.cs
```diff
@@ -100,9 +100,18 @@ public class ClearMethod
             [Fact]
             public void ShouldRemoveAdd()
```
lutovich: Maybe `ShouldClear`?

lutovich (approved):
@zhenlineo changed look good, one minor comment

PR #184, opened 2017-06-20
first draft of async api

ali-ince:
For the sake of traceability I introduced seperate interfaces which include async versions of their counterparts (IStatementRunnerAsync, ISessionAsync, ITransactionAsync and IStatementResultAsync) and make all implement their corresponding synchronous versions.

IDriver.Session methods are changed to return ISessionAsync that enables the user to call async version of methods.

We may later on decide whether these interfaces will remain in the public API or not. Please note that whatever we do we will need another IStatementResult interface that supports asynchronous operations and do not implement IEnumerable<IRecord>.

All interfaces introduced are implemented on their concrete classes and coded to call synchronous methods to perform their operations. This is just to see what we will feel while writing async programs with this draft API - optionally running what we write.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Driver.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Driver.cs
@@ -39,7 +39,7 @@ namespace Neo4j.Driver.Internal
             _retryLogic = retryLogic;
         }
 
-        public ISessionAsync Session(AccessMode defaultMode=AccessMode.Write, string bookmark = null)
+        public ISession Session(AccessMode defaultMode=AccessMode.Write, string bookmark = null)
         {
             return Session(defaultMode, Bookmark.From(bookmark, _logger));
         }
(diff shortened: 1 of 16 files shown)


ali-ince:
Some usage scenerios could be found as tests in Neo4j.Driver.IntegrationTetsts\ExamplesAsync.cs file - with same scenerios from Examples.cs file.

Neo4j.Driver/Neo4j.Driver/Internal/Driver.cs
```diff
@@ -39,7 +39,7 @@ internal Driver(Uri uri, IConnectionProvider connectionProvider, IRetryLogic ret
             _retryLogic = retryLogic;
         }
 
-        public ISession Session(AccessMode defaultMode=AccessMode.Write, string bookmark = null)
+        public ISessionAsync Session(AccessMode defaultMode=AccessMode.Write, string bookmark = null)
```
zhenlineo: Shall we still have `ISession`, `ITransaction`, and `IStatementResult` as exposed types? I feel the surfix `Aync` is not very typical to used in interfaces. What do you think?

zhenlineo (commented):
It looks really good.

One special comment: 
What does `Session.Dispose` and `Transaction.Dispose` mean to async API?

The current synchronized impl uses `Dispose` to return the connection assigned to session/tx back to the connection pool. While returning a connection back to the pool, all un-pulled result will be buffered into memory (a.k.a. Read all records) and then a `RESET` message will be sent to clean the connection state (a.k.a. Write RESET and Read reply of RESET).

In other words, the current `Session.Dispose` and `Transaction.Dispose` is doing some work that touches its underlying socket. `using` with resources does not fly with async very well. Therefore I am thinking if we need to abandon `using` pattern and introduce `CloseAsync` for the async API.

Neo4j.Driver/Neo4j.Driver/Internal/Driver.cs
ali-ince: You're certainly right about the ```Dispose``` point. Dispose is the facilitator of ```using``` block but more importantly is also called by the garbage collector to ensure every related resource is also disposed.

It is not arguable that most of our API contracts are designed to be used inside ```using``` blocks but there may be some use cases where it will not be an option. For this, I believe that we have to introduce Close & CloseAsync methods to required interfaces and call those from inside ```Dispose```.

We can talk about the naming alternatives and the way to introduce these async stuff into the API.

PR #185, opened 2017-06-27
introduce async capable methods on existing classes

ali-ince:
TcpSocketClient/SocketClient changes for async api

1. Changed ITcpSocketClient#DisconnectAsync to ITcpSocketClient#Disconnect since no async stuff inside.
2. Added timeOut parameter to ITcpSocketClient#ConnectAsync to better handle time outs across connection retries.
3. Modified TcpSocketClient / SocketClient and affected tests to honour these changes.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
@@ -18,6 +18,7 @@ using System;
 using System.IO;
 using System.Net.Security;
 using System.Net.Sockets;
+using System.Threading;
 using System.Threading.Tasks;
 using Neo4j.Driver.V1;
 using static System.Security.Authentication.SslProtocols;
@@ -51,14 +52,14 @@ namespace Neo4j.Driver.Internal.Connector
         public Stream ReadStream => _stream;
         public Stream WriteStream => _stream;
 
-        public async Task DisconnectAsync()
+        public void Disconnect()
         {
             Close();
         }
 
-        public async Task ConnectAsync(Uri uri)
+        public async Task ConnectAsync(Uri uri, TimeSpan timeOut)
         {
-            await Connect(uri);
+            await Connect(uri, timeOut);
             if (!_encryptionManager.UseTls)
             {
                 _stream = _client.GetStream();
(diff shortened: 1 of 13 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/TcpSocketClient.cs
```diff
@@ -81,27 +82,32 @@ public TcpSocketClient(EncryptionManager encryptionManager, bool keepAlive, bool
...
-                catch (Exception e)
+                var addresses = await uri.ResolveAsync(_ipv6Enabled);
+                AggregateException innerErrors = null;
+                for (var i = 0; i < addresses.Length; i++)
                 {
-                    var error = new IOException($"Failed to connect to server '{uri}' via IP address '{addresses[i]}': {e.Message}", e);
-                    innerErrors = innerErrors == null ? new AggregateException(error) : new AggregateException(innerErrors, error);
+                    try
+                    {
+                        cancellationSource.Token.ThrowIfCancellationRequested();
```
zhenlineo, line 95: If I understand this correctly, this timeout newly added is not for timing out `_client.ConnectAsync` but for this `for(var i = 0; i < address.Length; i++)`.

My question is if this timeout is necessary? Should we consider having timeout on `_client.ConnectAsync` instead?

ali-ince: I have done that change to remove Wait() call on [this line](https://github.com/neo4j/neo4j-dotnet-driver/blob/1.5/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs#L70) and move the timeout logic to a more manageable level - without blocking.

ali-ince: Before making any changes to the current behaviour, we should clarify what we will do when a single socket connection times out in the for-loop. Do we cancel all the connection process (throw an exception) or will we continue trying the next IPAddress available? 

It depends on how "establishing a connection" is conceived on the developer side. Is it the maximum time allowed to get a working SocketConnection or is it the maximum time to connect to an IPAddress as a pure Socket - which on failure an actual working SocketConnection could take some multitude of the timeout value given.

Keeping it as it is for not introducing a behavioral change. Will work on it when we clarify its behavior-to-be.

PR #190, opened 2017-06-29
Async methods in ConnectionPool

zhenlineo:
Removed ambiguous `IConnection.Dispose` method as it means differently for different `IConnection` implementation:

* `SocketConneciton` - Close network connection directly.
* PooledConnection - Reset connection and Release connection back to pool. Requires network send and receive.
* `TransactionConnection` - Release `Delegated` connection
* `ClusterConnection` - Perform `Delegated.Dispose()`

Instead `IConnection.Close` and `IConnection.CloseAsync()` are added as replacement to `IConnection.Dispose`
Renamed old`IConnection.Close` to `IConnection.Destroy` to better describe the only action in this method is to close/destroy resources only directly in this method.
Removed some unnecessary async method in `SocketClient` and `TcpClient`

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
@@ -66,7 +66,7 @@ namespace Neo4j.Driver.Internal.Connector
 
         public void Init()
         {
-            var connected = Task.Run(() => _client.Start(_connectionTimeout)).Wait(_connectionTimeout);
+            var connected = Task.Run(() => _client.StartAsync(_connectionTimeout)).Wait(_connectionTimeout);
             if (!connected)
             {
                 throw new IOException(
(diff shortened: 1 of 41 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -225,6 +216,12 @@ protected virtual void Dispose(bool isDisposing)
             }
         }
 
+        public Task CloseAsync()
+        {
+            Close();
+            return Task.CompletedTask; // TODO verify this is the correct way to do it
```
zhenlineo, line 230: @ali-ince What is the best way to impl a async method that actually do not need to wait for anything?
As you could see, this CloseAsync() on SocketConnection is the same as Close()

Neo4j.Driver/Neo4j.Driver/Internal/Transaction.cs
```diff
@@ -223,13 +223,19 @@ public TransactionConnection(Transaction transaction, IConnection connection)
...
             {
                 // no resouce will be closed as the resources passed in this class are managed outside this class
                 Delegate = null;
                 _transaction = null;
             }
 
+            public override Task CloseAsync()
+            {
+                Close();
+                return Task.CompletedTask;
```
zhenlineo, line 234: Same here

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
ali-ince: @zhenlineo I thought to remove those Async methods on our second pass - as we really do not need them at all. Instead we should use Close().

PR #191, opened 2017-06-29
Async StatementResultReader

ali-ince:
1. renamed IStatementResultAsync to IStatementResultReader,
2. implemented this new async interface,
3. replaced temporary async implementations in Session & Transaction with it.


zhenlineo:
I am merging this now as it looks very exciting that I might could try to plugin ConnectionPool async too.

PR #193, opened 2017-07-03
Improve tests & result builder classes

ali-ince:
1. Added tests for Session#RunAsync, Session#BeginTransactionAsync and Session#Dispose after async calls,
2. Added tests for Transaction#RunAsync,
3. Improved ResultBuilder & ResultReaderBuilder with a new base class including common members from both classes,
4. Added tests for ResultReaderBuilder & StatementResultReader,
5. Added Session#BeginTransactionAsync(string bookmark) method.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultBuilder.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultBuilder.cs
@@ -22,26 +22,20 @@ using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal.Result
 {
-    internal class ResultBuilder : IMessageResponseCollector
+    internal class ResultBuilder : BuilderBase
     {
-        private readonly List<string> _keys = new List<string>();
-        private readonly SummaryBuilder _summaryBuilder;
-
         private Action _receiveOneAction;
 
         private readonly Queue<IRecord> _records = new Queue<IRecord>();
         private bool _hasMoreRecords = true;
-
-        private readonly IResultResourceHandler _resourceHandler;
 
         public ResultBuilder() : this(null, null, null, null, null)
         {
         }
 
         public ResultBuilder(Statement statement, Action receiveOneAction, IServerInfo server, IResultResourceHandler resourceHandler = null)
+            : base(statement, server, resourceHandler)
         {
-            _summaryBuilder = new SummaryBuilder(statement, server);
-            _resourceHandler = resourceHandler;
             SetReceiveOneAction(receiveOneAction);
         }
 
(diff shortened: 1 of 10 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -427,6 +427,13 @@ public Task<ITransactionAsync> BeginTransactionAsync()
             return TryExecuteAsync(async()=> await BeginTransactionWithoutLoggingAsync(_defaultMode));
         }
 
+        public Task<ITransactionAsync> BeginTransactionAsync(string bookmark)
```
zhenlineo: There should be no BeginTx with bookmark as it is deprecated method already

Neo4j.Driver/Neo4j.Driver/Internal/Result/StatementResultReader.cs
```diff
@@ -32,9 +32,16 @@ public StatementResultReader(List<string> keys, Func<Task<IRecord>> nextRecordFu
...
+            if (_summary == null)
             {
-                _summary = _summaryFunc();
+                if (_summaryFunc != null)
+                {
+                    _summary = _summaryFunc();
+                }
+                else
+                {
+                    _summary = Task.FromResult((IResultSummary)null);
```
zhenlineo: Why this change is needed? Is there a test failure before?

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultBuilder.cs
```diff
@@ -22,26 +22,20 @@
...
-        private readonly SummaryBuilder _summaryBuilder;
-
         private Action _receiveOneAction;
 
         private readonly Queue<IRecord> _records = new Queue<IRecord>();
         private bool _hasMoreRecords = true;
 
-        private readonly IResultResourceHandler _resourceHandler;
-
         public ResultBuilder() : this(null, null, null, null, null)
```
zhenlineo: Why this constructor is allowed?

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultReaderBuilder.cs
```diff
@@ -23,23 +23,20 @@
...
+    internal class ResultReaderBuilder : BuilderBase
     {
-        private readonly List<string> _keys = new List<string>();
-        private readonly SummaryBuilder _summaryBuilder;
-
         private Func<Task> _receiveOneFunc;
 
         private readonly Queue<IRecord> _records = new Queue<IRecord>();
         private bool _hasMoreRecords = true;
+        public ResultReaderBuilder() : this(null, null, null, null, null)
```
zhenlineo: Same here. Why we need this constructor? If we will never use this in our code, won't it be better if we have no such a method to protect from using this class wrongly?

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultReaderBuilder.cs
```diff
@@ -23,23 +23,20 @@
 
 namespace Neo4j.Driver.Internal.Result
 {
-    internal class ResultReaderBuilder : IMessageResponseCollector
+    internal class ResultReaderBuilder : BuilderBase
```
zhenlineo, line 26: Is it easy to break the logic of `IMessageResponseCollector` to an independent class such as `MessageResponseCollector` and have an field in this class instead?
Suggesting to change in another PR if agreed.

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultReaderBuilder.cs
```diff
@@ -85,7 +82,7 @@ public IStatementResultReader PreBuild()
             return _records.Count > 0 ? _records.Dequeue() : null;
         }
 
-        private void SetReceiveOneFunc(Func<Task> receiveOneAction)
+        internal void SetReceiveOneAction(Func<Task> receiveOneAction)
```
zhenlineo: About naming: 
It is actually a function, so suggesting to name this to `SetReceiveOneFunc`

zhenlineo (commented):
A few naming comments.

PR #195, opened 2017-07-05
Allow session and transaction to be closed multiple times.

zhenlineo:
Allow DisposeAsync and Dispose to be called and not block on Dispose after DisposeAsync.


ali-ince:
Shall we also include a test that ensures upon a Dispose call after an async method call closes up everything as expected?

Neo4j.Driver/Neo4j.Driver.Tests/TransactionTests.cs
```diff
@@ -28,15 +31,43 @@ namespace Neo4j.Driver.Tests
 {
     public class TransactionTests
     {
+        private static Expression<Action<IConnection>> RunRollback => x => x.Run("ROLLBACK", null, null, false);
```
ali-ince, line 34: The previous usage pattern was a bit more self describing about what is expected at each call, but we may give a try anyways.

zhenlineo:
Done :)

PR #199, opened 2017-07-13
Update nuget dependency to latest

zhenlineo:


ali-ince (approved):
I also checked locally and everything was fine!

PR #200, opened 2017-07-17
Decouple routing logic from address set

lutovich:


Neo4j.Driver/Neo4j.Driver/Internal/Routing/AddressSet.cs
```diff
@@ -14,136 +14,105 @@
...
-                {
-                    value = default(T);
-                    return false;
-                }
-                // ensure the index is in range
-                _index = _index % _items.Count;
-                value = _items[_index++];
-            }
-            return true;
+            _items = new List<T>();
```
ali-ince: Should we also make this assignment a critical section?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/IAddressSet.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Not thread safe
+        /// </summary>
+        bool IsEmpty { get; }
+
+        /// <summary>
+        /// Not thread safe
+        /// </summary>
+        /// <returns></returns>
+        string ToString();
```
ali-ince: There's no need for a ```ToString()``` declaration as it is part of ```Object``` class that can be overrided.

ali-ince (approved):
Your changes looks ok. A few comments only.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/AddressSet.cs
lutovich: Yes, good point. It is fine for `#Clear()` to assign because it is an atomic action, but it is not fine for it to not coordinate with other modification methods like `#Add()` or `#Remove()`.

lutovich:
@ali-ince comments should now be addresses. Thank you.

PR #201, opened 2017-07-18
Least connected load balancing

lutovich:
This PR introduces least connected load balancing strategy and makes it default. Experimental configuration option is added to change it back to round-robin.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Routing/LeastConnectedLoadBalancingStrategy.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Routing/LeastConnectedLoadBalancingStrategy.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal.Routing
 {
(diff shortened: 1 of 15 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Routing/LeastConnectedLoadBalancingStrategy.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            var count = addresses.Count;
+            if (count == 0)
+            {
+                _logger.Trace($"Unable to select {addressType}, no known addresses given");
+                return null;
+            }
+
+            // choose start index for iteration in round-rodin fashion
+            var startIndex = roundRobinIndex.Next(count);
```
ali-ince, line 58: Does this provide any kind of advantage over a 0..count-1 loop? Either way, we are iterating through all of the list items and selecting the least connected one. Anything I could not see?

ali-ince (approved):
Looks ok.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LeastConnectedLoadBalancingStrategy.cs
lutovich: The point here is to "randomize" address selection when all addresses have same amount of active connections. This is especially valuable when cluster has no active transactions. With simple for-loop we would always select the first available address. With round-robin selection of starting point we select machines in round-robin fashion. It could save us when accidentally fist machine is much worse than others.

ali-ince: Yeah, that's right. Need some sleep I guess :)

PR #203, opened 2017-07-25
Add async support for routing driver

zhenlineo:
Improved statistic collector to be able to pull data from multiple statistic provider and remove statistic provider once it is no longer valid.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
@@ -465,8 +465,8 @@ namespace Neo4j.Driver.Internal
         {
             if (_statistics != null)
             {
-                _statistics.Dispose();
                 _statisticsCollector?.Unregister(_statistics);
+                _statistics.Dispose();
             }
         }
 
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
@@ -47,7 +47,8 @@ namespace Neo4j.Driver.Internal
         // for test only
         private readonly IConnection _fakeConnection;
 
-        private readonly ConnectionPoolStatistics _statistics;
+        private IStatisticsCollector _statisticsCollector;
+        private ConnectionPoolStatistics _statistics;
 
         public int NumberOfInUseConnections => _inUseConnections.Count;
         internal int NumberOfAvailableConnections => _availableConnections.Count;
@@ -73,12 +74,7 @@ namespace Neo4j.Driver.Internal
 
             _logger = logger;
 
-            var statisticsCollector = connectionPoolSettings.StatisticsCollector;
-            if (statisticsCollector != null)
-            {
-                _statistics = new ConnectionPoolStatistics(uri, this);
-                statisticsCollector.Register(_statistics);
-            }
+            SetupStatisticsProvider(connectionPoolSettings.StatisticsCollector);
         }
 
         internal ConnectionPool(
@@ -446,13 +442,32 @@ namespace Neo4j.Driver.Internal
                     DestoryConnection(connection);
                 }
             });
-            _statistics?.Dispose();
+            DisposeStatisticsProvider();
             base.Dispose(true);
         }
 
         private void ThrowObjectDisposedException()
         {
             FailedToCreateConnection(this);
+        }
+
+        private void SetupStatisticsProvider(IStatisticsCollector collector)
+        {
+            _statisticsCollector = collector;
+            if (_statisticsCollector != null)
+            {
+                _statistics = new ConnectionPoolStatistics(_uri, this);
+                _statisticsCollector.Register(_statistics);
+            }
+        }
+
+        private void DisposeStatisticsProvider()
+        {
+            if (_statistics != null)
+            {
+                _statistics.Dispose();
+                _statisticsCollector?.Unregister(_statistics);
+            }
         }
 
         public override string ToString()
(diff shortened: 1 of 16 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -455,6 +451,25 @@ private void ThrowObjectDisposedException()
...
+                _statisticsCollector.Register(_statistics);
+            }
+        }
+
+        private void DisposeStatisticsProvider()
+        {
+            if (_statistics != null)
+            {
+                _statistics.Dispose();
+                _statisticsCollector?.Unregister(_statistics);
```
ali-ince: Though it does not make any difference specifically with this usage scenerio, these two lines should be swapped for avoiding access to a disposed object.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPoolStatistics.cs
```diff
@@ -81,6 +94,7 @@ public void Dispose()
             _pool = null;
         }
 
-        public static ConnectionPoolStatistics Read(IDictionary<string, object> dict) => new ConnectionPoolStatistics(dict);
+        public static ConnectionPoolStatistics Read(string name, IDictionary<string, object> dict)
```
ali-ince: Better to have it named as From or FromDictionary?

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/CollectionExtensions.cs
```diff
@@ -49,7 +49,7 @@ public static T GetValue<T>(this IDictionary<string, object> dict, string key, T
 
         public static string ToContentString<K, V>(this IDictionary<K, V> dict)
         {
-            var output = dict.Select(item => $"{{{item.Key}, {item.Value}}}");
+            var output = dict.Select(item => $"{{{item.Key}, {item.Value.ValueToString()}}}");
```
ali-ince, line 52: As an improvement, I think we should develop more readable display format (may be indented?) for statistics as it becomes hard to decompose & understand especially when output is large.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnectionPool.cs
```diff
@@ -79,6 +80,18 @@ public IConnection Acquire(Uri uri)
             return pool.Acquire(ignored);
         }
 
+        public async Task<IConnection> AcquireAsync(Uri uri)
```
ali-ince: async and await keywords could be avoided in this method.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscoveryManager.cs
```diff
@@ -61,46 +61,95 @@ public void Rediscovery()
...
-                    }
-                    ExpireAfterSeconds = record["ttl"].As<long>();
+        public async Task RediscoveryAsync()
+        {
+            var provider = new SingleConnectionBasedConnectionProvider(_conn);
+            var session = new Session(provider, _logger);
+            try
+            {
+                var result = await session.RunAsync(DiscoveryProcedure).ConfigureAwait(false);
+                await result.ReadAsync().ConfigureAwait(false);
```
ali-ince: It would be better to standardize this usage as 
```csharp
if (await result.ReadAsync().ConfigureAwait(false)) {

} else {
   throw ....
}
```
or negation of it.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscoveryManager.cs
```diff
@@ -130,7 +179,17 @@ public IConnection Acquire(AccessMode mode)
 
             public Task<IConnection> AcquireAsync(AccessMode mode)
             {
-                throw new NotImplementedException();
+                var task = new TaskCompletionSource<IConnection>();
```
ali-ince: This can be simplified as

```csharp
return Task.FromResult(Acquire(mode));
```

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscoveryManager.cs
```diff
@@ -130,7 +179,17 @@ public IConnection Acquire(AccessMode mode)
...
+                task.SetResult(Acquire(mode));
+                return task.Task;
+            }
+
+            public async Task CloseAsync()
+            {
+                if (_connection != null)
```
ali-ince: async & await keywords could be avoided in this method, like

```csharp
if (_connection != null) {
   return _connection.CloseAsync();
}

return Task.CompletedTask;
```

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/CollectionExtensions.cs
zhenlineo: Statistics will be reported in a dictionary. It depends on how the end user would like to print. 

The real way of consuming statistics in my mind was that we will give the dictionary to a graphic server and show us all statistics in real time (every a few sec). I would suggest to not bother too much of how we internally print them out now. We are not really relying on console printout in our tests anyway.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscoveryManager.cs
zhenlineo: Actually this place
```
await result.ReadAsync().ConfigureAwait(false);
var record = result.Current();
```

What we really want is
```
await result.SingleAsync().ConfigureAwait(false);
```
So I added an extension class to impl this `SingleAsync` now :D

PR #205, opened 2017-07-29
refactor of packstream & bolt layer

ali-ince:
Refactored packstream & bolt layer to better handle async logic


Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketClient.cs
```diff
@@ -121,24 +161,24 @@ public void Send(IEnumerable<IRequestMessage> messages)
...
-                await _writer.FlushAsync().ConfigureAwait(false);
             }
             catch (Exception ex)
             {
                 _logger?.Info($"Unable to send message to server {_uri}, connection will be terminated. ", ex);
                 Stop();
                 throw;
             }
+
+            return _writer.FlushAsync();
```
zhenlineo: This changes the original flow. The `catch` cause was aiming to catch any error to flush.

Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        public Task ReadAsync(IMessageResponseHandler responseHandler)
+        {
+            _bufferStream.SetLength(0);
+
+            return
+                _chunkReader.ReadNextChunkAsync(_bufferStream)
+                    .ContinueWith(t =>
+                    {
```
zhenlineo: Maybe we could abstract the handling of the structure into a separate method that both async and sync method could call?

Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
```diff
@@ -0,0 +1,279 @@
...
+            _packStreamReader = supportBytes ? new PackStreamReader(_bufferStream) : new PackStreamReaderBytesIncompatible(_bufferStream);
+        }
+
+
+
+        public void Read(IMessageResponseHandler responseHandler)
+        {
+            _bufferStream.SetLength(0);
+
+            _chunkReader.ReadNextChunk(_bufferStream);
```
zhenlineo: If I got the code here correctly, the whole logic depends on the assumption that a whole message (such as a Record message) could be put into `_bufferStream`. I do not think that we could assume on that.

Pls correct me if I understand this part wrongly.

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
```diff
@@ -0,0 +1,236 @@
...
+            {
+                // We have not received the chunk size yet, but it can be read from the buffered data.
+                if (_currentChunkSize == -1 && HasBytesAvailable(_chunkSizeBuffer.Length))
+                {
+                    ReadFromChunkStream(_chunkSizeBuffer, 0, _chunkSizeBuffer.Length);
+
+                    _currentChunkSize = PackStreamBitConverter.ToUInt16(_chunkSizeBuffer);
+                    if (_currentChunkSize == 0)
+                    {
+                        break;
```
zhenlineo, line 74: So this while loop will not stop until we've reached a `0x00 00`, does this mean that we will not stop putting more data into `targetStream` (a.k.a. `boltReader._bufferStream`) until we pulled the whole message into `targetStream`?

As I committed in `boltReader` I do not think we could assume that we could always pull the whole message into memory. CC @technige I need some inputs from you too.

Similar for writer, when a chunk is finished, then we should flush it immediately.

zhenlineo (changes requested):
Have not finished the whole PR, but here are some key notes that we need to clarify before we continue.

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketClient.cs
ali-ince: Fixed :)

Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
ali-ince: Done :)

zhenlineo (approved):

PR #211, opened 2017-08-04
Enriched the extension helpers on async result

zhenlineo:
Added a few extension methods for handling async result, such as SingleAsync, ForeachAsync.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/ExamplesAsync.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/ExamplesAsync.cs
@@ -303,10 +303,7 @@ namespace Neo4j.Driver.ExamplesAsync
                 {
                     var result = await tx.RunAsync("SELECT * FROM Employees WHERE name = $name", new { name });
 
-                    if (await result.ReadAsync())
-                        return result.Current()["employee_number"].As<int>();
-                    else
-                        return 0;
+                    return (await result.SingleAsync())["employee_number"].As<int>();
                 }
                 catch (ClientException ex)
                 {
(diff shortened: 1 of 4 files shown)


ali-ince (approved):

zhenlineo:
@ali-ince I removed the extension method 
`Task ForEachAsync(..., Function<IRecord, Task>)`
but instead added 
`Task<List<T>> ToListAsync<T>(..., Function<IRecord, T>)`

The reason I changed is that I am not sure if we should make decisions for a user to await on their tasks. Instead I think it might be better to give `ToListAsync` method where a user could return his/her tasks into a `list<Task>` and then waitAll on them.

And in the future, when requested, we could add 
`Task ForEachAsync(..., Function<IRecord, Task>)` and
`Task ForEachAsync<T>(..., Function<IRecord, Task<T>>)`

Neo4j.Driver/Neo4j.Driver.IntegrationTests/ExamplesAsync.cs
```diff
@@ -509,12 +501,7 @@ public ResultConsumeExample(ITestOutputHelper output, StandAloneIntegrationTestF
                     {
                         var result = await tx.RunAsync("MATCH (a:Person) RETURN a.name ORDER BY a.name");
 
-                        List<string> list = new List<string>();
-
-                        while (await result.ReadAsync())
-                            list.Add(result.Current()[0].As<string>());
-
-                        return list;
+                        return await result.ToListAsync(r => r[0].As<string>());
```
zhenlineo, line 504: This code shows a "normal" usecase of `Task<List<T>> ToListAsync(..., Func<IRecord, T>)`

PR #213, opened 2017-08-08
modified CollectionExtensions.ToDictionary to handle nested objects

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/Extensions/CollectionExtensions.cs
```diff
@@ -86,8 +88,46 @@ public static string ValueToString(this object o)
...
+        {
+            foreach (var propInfo in o.GetType().GetRuntimeProperties())
+            {
+                var name = propInfo.Name;
+                var value = propInfo.GetValue(o);
+
+                if (value != null)
+                {
+                    var convertible = value as IConvertible;
+                    if (convertible != null)
```
zhenlineo: What would happen if `covertible` is a dictionary or list?
Could you add some tests for these two situations?

ali-ince: Yeah, missed those conditions. What about the members of arrays, lists and dictionaries? Shall we also handle conversion of their elements?

zhenlineo: Ideally it should be a recursive call, have a look at [java driver's impl](https://github.com/neo4j/neo4j-java-driver/blob/1.5/driver/src/main/java/org/neo4j/driver/v1/Values.java#L68)

Each elements of list is a recursive call [here](https://github.com/neo4j/neo4j-java-driver/blob/1.5/driver/src/main/java/org/neo4j/driver/v1/Values.java#L195)

And according to [this](https://github.com/neo4j/neo4j-dotnet-driver/blob/1.5/Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamWriter.cs#L40), we have to convert array to IList too

ali-ince: Introduced recursive conversion of objects in the object hierarchy - supporting arrays, lists and dictionaries. Also added a number of integration tests that test against possible use cases documented [here](http://neo4j.com/docs/developer-manual/current/cypher/syntax/parameters/).

PR #215, opened 2017-08-09
Change from optional methods to method overloading

zhenlineo:
Made API class to have method overloading without changing the original method meaning.


ali-ince (approved):
changes look fine!

PR #217, opened 2017-08-10
Avoiding throw AggregateException

zhenlineo:
Made Retry and TcpClient to not throw AggregatedException on the top level. 
But as Retry and TcpClient still try to do some work for several times, the inner errors as a list are still stored in a AggregatedException.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/TransactionIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/TransactionIT.cs
@@ -46,11 +46,12 @@ namespace Neo4j.Driver.IntegrationTests
                 }));
                 timer.Stop();
 
-                var error = e as AggregateException;
+                var error = e.InnerException as AggregateException;
                 var innerErrors = error.Flatten().InnerExceptions;
                 foreach (var innerError in innerErrors)
                 {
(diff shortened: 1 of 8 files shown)


Neo4j.Driver/Neo4j.Driver.Tests/RetryLogicTests.cs
```diff
@@ -62,7 +62,7 @@ private void Retry(int index, IRetryLogic retryLogic)
             }));
             timer.Stop();
 
-            var error = e as AggregateException;
+            var error = e.InnerException as AggregateException;
```
ali-ince: We shall also check for e's type against ServiceUnavailableException.

Neo4j.Driver/Neo4j.Driver.IntegrationTests/DirectDriver/TransactionIT.cs
```diff
@@ -46,11 +46,12 @@ public void ShouldRetry()
                 }));
                 timer.Stop();
 
-                var error = e as AggregateException;
+                var error = e.InnerException as AggregateException;
```
ali-ince: We shall also check for e's type against ServiceUnavailableException.

ali-ince (changes requested):

zhenlineo:
@ali-ince Done

PR #218, opened 2017-08-14
1.5 async performance improvement

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
```diff
@@ -81,14 +81,22 @@ public Task ReadAsync(IMessageResponseHandler responseHandler)
...
+                        ConsumeMessages(responseHandler);
                     }, TaskContinuationOptions.ExecuteSynchronously);
         }
 
-        private void ProcessMessage(IMessageResponseHandler responseHandler)
+        private void ConsumeMessages(IMessageResponseHandler responseHandler)
         {
             _bufferStream.Position = 0;
 
+            while (_bufferStream.Length > _bufferStream.Position)
```
zhenlineo, line 92: I am a bit confused as I did not get why this will ever be false???
I thought _bufferStream (MemoryStream) is infinite, isn't it?

zhenlineo:
It looks to me that this PR is about pulling as much as messages that have arrived at the network input buffer into memory to reduce reading operations from the network.

If that the case, then it would be a good change if we have a fixed buffer size. Otherwise, for receiving records, we are just blocking on the whole result to arrive and then process them together. We then breaks the contract of streaming records.

Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
ali-ince: _chunkReader reads a finite number of bytes into this _bufferStream and its Length does not change in ConsumeMessages. However, while we're reading messages from it, we are advancing its Position. This loop ensures that all messages that are read in one ReadNextMessage(Async) call has been consumed.

ali-ince:
I am already doing that change and will make a PR later on.

However, the ChunkReader.ReadNextMessage(Async) calls trigger themselves until a complete message is read into buffers. The latter case of pulling further messages does not occur at the network level. It just looks into the buffers if we have somehow leftover data that can make (a) whole message(s). This is mostly probable for very small messages.

Neo4j.Driver/Neo4j.Driver/Internal/IO/BoltReader.cs
zhenlineo: I see, then I totally misunderstood this code. I will merge this when the build finished.

PR #223, opened 2017-08-16
MaxConnLifetime

zhenlineo:
Added maxConnectionLifeTime into config.
When this config option is set to a non-negative value, the connection will be closed rather than reused when it is obtained from the connection pool.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -665,8 +665,8 @@ namespace Neo4j.Driver.Tests
             public void ShouldReturnZeroAfterCreation()
             {
                 var uri = new Uri("localhost:7687");
-                var connectionSettings = new ConnectionSettings(uri, AuthTokens.None, Config.DefaultConfig);
-                var poolSettings = new ConnectionPoolSettings(1, 1, TimeSpan.MaxValue, TimeSpan.MaxValue);
+                var connectionSettings = new ConnectionSettings(AuthTokens.None, Config.DefaultConfig);
+                var poolSettings = new ConnectionPoolSettings(1, 1, Config.Infinite, Config.Infinite, Config.Infinite);
                 var bufferSettings = new BufferSettings(Config.DefaultConfig);
                 var logger = new Mock<ILogger>().Object;
 
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
@@ -6,12 +6,12 @@ namespace Neo4j.Driver.Internal
     internal class ConnectionValidator
     {
         private readonly TimeSpan _connIdleTimeout;
-        private readonly TimeSpan _maxConnLifeTime;
+        private readonly TimeSpan _maxConnLifetime;
 
-        public ConnectionValidator(TimeSpan connIdleTimeout, TimeSpan maxConnLifeTime)
+        public ConnectionValidator(TimeSpan connIdleTimeout, TimeSpan maxConnLifetime)
         {
             _connIdleTimeout = connIdleTimeout;
-            _maxConnLifeTime = maxConnLifeTime;
+            _maxConnLifetime = maxConnLifetime;
         }
 
         public bool IsConnectionReusable(IPooledConnection connection)
@@ -79,11 +79,11 @@ namespace Neo4j.Driver.Internal
         // I like this method name :P
         private bool HasBeenAliveForTooLong(IPooledConnection connection)
         {
-            if (!IsConnectionLifeTimeDetectionEnabled())
+            if (!IsConnectionLifetimeDetectionEnabled())
             {
                 return false;
             }
-            if (connection.LifeTimeTimer.ElapsedMilliseconds > _maxConnLifeTime.TotalMilliseconds)
+            if (connection.LifetimeTimer.ElapsedMilliseconds > _maxConnLifetime.TotalMilliseconds)
             {
                 return true;
             }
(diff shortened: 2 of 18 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly TimeSpan _connIdleTimeout;
+        private readonly TimeSpan _maxConnLifeTime;
+
+        public ConnectionValidator(TimeSpan connIdleTimeout, TimeSpan maxConnLifeTime)
+        {
+            _connIdleTimeout = connIdleTimeout;
+            _maxConnLifeTime = maxConnLifeTime;
+        }
+
+        public bool IsConnectionReusable(IPooledConnection connection)
```
ali-ince: Shall we change this method's name to be more self describing such as ResetAndRestartIdleTimer because it has side effects aside from checks.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            if (!connection.IsOpen)
+            {
+                return false;
+            }
+
+            try
+            {
+                connection.ClearConnection();
```
ali-ince, line 26: Not specific for this PR, but if in the middle of a huge resultset streaming, calling RESET may stop the server from streaming more data but it will not help client's consuming pending network buffers - which may take considerable time. Shall we have a means of timeout-ing?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                return false;
+            }
+
+            RestartIdleTimer(connection);
+
+            return true;
+        }
+
+        public async Task<bool> IsConnectionReusableAsync(IPooledConnection connection)
```
ali-ince, line 38: The same two comments applies for this method.

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
```diff
@@ -665,8 +665,8 @@ public class NumberOfInUseConnections
             public void ShouldReturnZeroAfterCreation()
             {
                 var uri = new Uri("localhost:7687");
-                var connectionSettings = new ConnectionSettings(uri, AuthTokens.None, Config.DefaultConfig);
-                var poolSettings = new ConnectionPoolSettings(1, 1, TimeSpan.MaxValue, TimeSpan.MaxValue);
+                var connectionSettings = new ConnectionSettings(AuthTokens.None, Config.DefaultConfig);
+                var poolSettings = new ConnectionPoolSettings(1, 1, Config.Infinite, Config.Infinite, Config.Infinite);
```
ali-ince, line 669: We may elaborate on what TimeSpan value makes it Infinite. I see the point that we accept -1 as an Infinite, but since we define the type of setting timeout values as TimeSpans, any TimeSpan value less than zero, zero itself or even TimeSpan.MaxValue could be perceived as infinite values. Not a big overhead when set something different than -1, but we shall be more natural.

ali-ince: Found System.Threading.TimeOut static class which declares InfiniteTimeSpan static field which is also using -1. No change needed then, but we may point out this in the docs.

zhenlineo: The current decision is that we treat all negative values as Infinite, which is stated in the API docs of `Config` for these `TimeSpan` settings.
The `Config.Infinite` value is `internal` and as you figured out I get this number from some standard lib.

ali-ince: Ok!

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
zhenlineo: Right now we do not provide any API method to stop streaming. Internally our current `sessoin.close` will also not stop any streaming but buffering all unconsumed results into memory. So we do not need to worry about timing out now.

But this is interesting when we want to add cancellation token into our async methods.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
zhenlineo: My thinking was that the outside of this object should not care about Idle timer or life time. The contract was that once this method returns true, then this connection can be reused, a.k.a. putting back to the connection pool. So I would prefer the name stays this `IsConnectionReusable`

ali-ince (approved):

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+            RestartIdleTimer(connection);
+
+            return true;
+        }
+
+        private void RestartIdleTimer(IPooledConnection connection)
+        {
+            if (IsConnectionIdleDetectionEnabled())
+            {
+                connection.IdleTimer.Start();
```
lutovich, line 63: I think it is a bit strange that validator resets timers. Can this happen in the connection pool instead?
Same happens in `HasBeenIdleForTooLong`.

Neo4j.Driver/Neo4j.Driver/V1/Config.cs
```diff
@@ -404,6 +416,15 @@ public interface IConfigBuilder
         IConfigBuilder WithConnectionIdleTimeout(TimeSpan timeSpan);
 
         /// <summary>
+        /// Specify the maximum connection life time.
+        /// The connection that has been created for longer than specified time will not be reused but closed.
+        /// </summary>
+        /// <param name="timeSpan">The max timespan that a connection can be reused after has been created for.</param>
+        /// <returns>An <see cref="IConfigBuilder"/> instance for further configuration options.</returns>
+        /// <remarks>Must call <see cref="ToConfig"/> to generate a <see cref="Config"/> instance.</remarks>
+        IConfigBuilder WithMaxConnectionLifeTime(TimeSpan timeSpan);
```
lutovich: Should publicly visible name be `Lifetime` and not `LifeTime`?

PR #224, opened 2017-08-16
 Improve buffer usage across bolt & packstream classes

ali-ince:
1. Introduced Buffer related settings into the configuration,
  2. Dropped MemoryStream from ChunkReader and started to use a fix sized buffer,
  3. Dropped MemoryStream from BoltWriter,
  4. Reset MemoryStream capacity to default buffer size after consuming or sending messages in BoltReader and ChunkWriter classes.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
@@ -26,13 +26,12 @@ namespace Neo4j.Driver.Internal.IO
     internal class ChunkReader: IChunkReader
     {
         private readonly Stream _downStream;
-        private readonly MemoryStream _chunkStream;
         private readonly ILogger _logger;
-        private long _lastWritePosition = 0;
-        private long _lastReadPosition = 0;
 
         private readonly byte[] _chunkSizeBuffer = new byte[2];
-        private readonly byte[] _buffer = new byte[8 * 1024];
+        private readonly byte[] _buffer = new byte[Constants.ChunkBufferSize];
+        private int _lastWritePosition = 0;
+        private int _lastReadPosition = 0;
         private int _currentChunkSize = -1;
 
         public ChunkReader(Stream downStream)
@@ -41,21 +40,12 @@ namespace Neo4j.Driver.Internal.IO
             
         }
 
-        public ChunkReader(Stream downStream, ILogger logger)
-            : this(downStream, new MemoryStream(), logger)
-        {
-            
-        }
-
-        internal ChunkReader(Stream downStream, MemoryStream chunkStream, ILogger logger)
+        internal ChunkReader(Stream downStream, ILogger logger)
         {
             Throw.ArgumentNullException.IfNull(downStream, nameof(downStream));
             Throw.ArgumentOutOfRangeException.IfFalse(downStream.CanRead, nameof(downStream));
 
-            Throw.ArgumentNullException.IfNull(chunkStream, nameof(chunkStream));
-
             _downStream = downStream;
-            _chunkStream = chunkStream;
             _logger = logger;
         }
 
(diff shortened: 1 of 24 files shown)


zhenlineo:
Looks okay, but would like to hear more input from @lutovich

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
```diff
@@ -41,33 +40,26 @@ public ChunkReader(Stream downStream)
...
 
         private bool TryReadFromBuffer(Stream targetStream)
         {
             while (true)
             {
+                // First try to retrieve the chunk size.
                 if (_currentChunkSize == -1 && HasBytesAvailable(_chunkSizeBuffer.Length))
                 {
-                    ReadFromChunkStream(_chunkSizeBuffer, 0, _chunkSizeBuffer.Length);
+                    ReadFromBuffer(_chunkSizeBuffer, 0, _chunkSizeBuffer.Length);
```
zhenlineo: This method is only used for reading chunk size, maybe we could just move all logic of reading size header into the same method and rename it as `ReadChunkSize()`?

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
```diff
@@ -76,47 +68,93 @@ private bool TryReadFromBuffer(Stream targetStream)
...
 
-                if (_currentChunkSize != -1 && HasBytesAvailable(_currentChunkSize + _chunkSizeBuffer.Length))
+                // As long as we know the chunk size and some bytes available in the buffers, write those
+                // to the target stream.
+                if (_currentChunkSize != -1 && HasBytesAvailable())
                 {
-                    CopyToFromChunkStream(targetStream, _currentChunkSize);
+                    var count = Math.Min(_currentChunkSize, _lastWritePosition - _lastReadPosition);
+
+                    CopyFromBuffer(targetStream, count);
```
zhenlineo: Will it be a good idea if we rename `targetStream` to `messageStream`/`dechunkedStream` and `downStream` to `networkStream`?

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
```diff
@@ -76,47 +68,93 @@ private bool TryReadFromBuffer(Stream targetStream)
...
-                    break;
+                    // Read next available bytes from the down stream and process it.
+                    var read = _downStream.Read(_buffer, _lastWritePosition, _buffer.Length - _lastWritePosition);
+
+                    _logger?.Trace("S: ", _buffer, _lastWritePosition, read);
+
+                    _lastWritePosition += read;
+
+                    // Can we read a whole message from what we have?
+                    if (TryReadFromBuffer(targetStream))
```
zhenlineo: `TryReadOneCompleteMessage`?

zhenlineo:
Reviewed with @lutovich, I got a few naming suggestions. He got the question if it is needed to expose the `defaultReadBufferSize`, `MaxReadBufferSize` and the same for writer exposed in Config? Shall we first hides it from users as it might be really hard to config them correctly by an end-user.

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
ali-ince: renamed to ```TryReadOneCompleteMessageFromBuffer```.

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
ali-ince: Renamed ```targetStream``` to ```messageStream``` but did not touch ```downStream``` as it may not be networkStream which we're dealing with.

Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs
ali-ince: done!

ali-ince:
Regarding config options, there are defaults set and unless the user does not have any specific needs (very large reads or very large writes) we do not expect them to set those config options. For me it is better to have config options with suitable defaults and let the user change them upon any requirement. We may discuss further if needed.

PR #228, opened 2017-08-22
Added MaxConnectionPoolSize support to sync method

zhenlineo:
Change the default value of MaxConnectionPoolSize to infinite

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -77,22 +77,22 @@ namespace Neo4j.Driver.Tests
                 pool.NumberOfAvailableConnections.Should().Be(0);
                 pool.NumberOfInUseConnections.Should().Be(2);
 
-                var timmer = new Stopwatch();
+                var timer = new Stopwatch();
                 var blockingAcquire = new Task(()=>{ pool.Acquire(); });
                 var releaseConn = new Task(()=>{ conn.Close(); });
 
-                timmer.Start();
+                timer.Start();
                 blockingAcquire.Start();
                 Task.Delay(1000).Wait(); // delay a bit here
                 releaseConn.Start();
 
                 releaseConn.Wait();
                 blockingAcquire.Wait();
-                timmer.Stop();
+                timer.Stop();
 
                 pool.NumberOfAvailableConnections.Should().Be(0);
                 pool.NumberOfInUseConnections.Should().Be(2);
-                timmer.ElapsedMilliseconds.Should().BeGreaterOrEqualTo(1000);
+                timer.ElapsedMilliseconds.Should().BeGreaterOrEqualTo(1000);
             }
 
             [Fact]
@@ -706,7 +706,7 @@ namespace Neo4j.Driver.Tests
             {
                 var uri = new Uri("localhost:7687");
                 var connectionSettings = new ConnectionSettings(AuthTokens.None, Config.DefaultConfig);
-                var poolSettings = new ConnectionPoolSettings(1, 1, Config.Infinite, Config.Infinite, Config.Infinite);
+                var poolSettings = new ConnectionPoolSettings(1, 1, Config.InfiniteInterval, Config.InfiniteInterval, Config.InfiniteInterval);
                 var bufferSettings = new BufferSettings(Config.DefaultConfig);
                 var logger = new Mock<ILogger>().Object;
 
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -18,6 +18,9 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices.ComTypes;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Moq;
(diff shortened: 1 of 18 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -284,7 +310,7 @@ public void Release(IPooledConnection connection)
                 }
                 if (_connectionValidator.IsConnectionReusable(connection))
                 {
-                    if (IsPoolFull())
+                    if (IsIdlePoolFull())
```
ali-ince, line 318: Checking IdlePoolSize upon Release calls would cause most of the connections to be destroyed with the default value of 10 given the MaxPoolSize is set to Infinite. I think that we shall either check against MaxPoolSize here or change the default value of 10.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -324,7 +350,7 @@ public Task ReleaseAsync(IPooledConnection connection)
 
                 if (await _connectionValidator.IsConnectionReusableAsync(connection).ConfigureAwait(false))
                 {
-                    if (IsPoolFull())
+                    if (IsIdlePoolFull())
```
ali-ince, line 358: Same here.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -33,9 +33,11 @@ internal class ConnectionPool : LoggerBase, IConnectionPool
         private readonly int _maxPoolSize;
         private readonly int _idlePoolSize;
         private readonly TimeSpan _connAcquisitionTimeout;
-        private readonly ConnectionValidator _connectionValidator;
+        internal TimeSpan ConnAcquisitionBlockingTimeout { get; set; } = TimeSpan.FromSeconds(5);
```
ali-ince: Do we need to introduce a new internal timeout setting that could not be configured by the end user? I think that we should either favor ConnectionAcquisitionTimeOut value instead of this or fine tune default values (ConnectionAcquisitionTimeOut is set as 1m whereas this is 5m). What do you think?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -185,9 +186,30 @@ public IPooledConnection Acquire()
 
                 if (!_availableConnections.TryTake(out connection))
                 {
-                    connection = CreateNewPooledConnection();
+                    var connAcquisitionTimer = new Stopwatch();
```
ali-ince, line 194: We may come up with a solution by incorporating CancellationTokens inside this code and make the timeout value to be applied across recursive calls. This behaviour makes more sense to me, what about you? If you agree, I can work on another PR after this one's approval.

ali-ince (commented):
Changes looks ok, I have commented on a few points though.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Do we need to make it recursive?
I thought`acquisition timeout` measures the waiting time to acquire a connection from pool each time. If we enter some recursive call at line 215, then it is another new timeout?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Here is 5s, which is the each small blocking timeout. So the idea was that when we blocking on waiting for a connection available from pool (here is 1m), there are two ways that we could claim a connection from pool: 1) waiting for pool to give us connection that is newly returned to the pool, 2) create a new one. This 5s timeout applies for how long we will block on 1) each time. In totally we will still blocking for 1mins, just we break every 5s that we will check if we could either do 2) or 1)

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
ali-ince: From user's point of view, if I set a connection acquisition timeout value I want it to be applied to the whole process whatever it does inside. At least I think that way :)

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
ali-ince: I totally read the defaults wrong, sorry. But in the end, if the user sets the ConnectionAcquisitionTime to something around 0-10 seconds, I guess this internal configured value can cause a problem.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: But`MaxPoolSize = infinite, idlePoolSize = 10` was not changed. This is what the old driver was using. Also the same behaviour for old users.

I wrote in the Config settings that it is highly suggested to not use poolSize = Infinite when using async API.

ali-ince: Ok.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: I will leave the final decision of this setting to you and @lutovich and/or anyone else who might be interested.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: I did a silly work around for your point that `ConnAcquisitionBlockingTimeout` will be reset to `AcquisitionTimeout` if `ConnAcquisitionBlockingTimeout > AcquisitionTimeout`. Pls also confirm with others what would be the best choice here.

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
```diff
@@ -65,17 +68,54 @@ public void ShouldCallConnInit()
...
             {
-                var connectionPoolSettings = new ConnectionPoolSettings(new Config {MaxIdleConnectionPoolSize = 2});
+                var connectionPoolSettings = new ConnectionPoolSettings(new Config {MaxConnectionPoolSize = 2});
                 var pool = new ConnectionPool(MockedConnection, settings: connectionPoolSettings);
+                var conn = pool.Acquire();
                 pool.Acquire();
+                pool.NumberOfAvailableConnections.Should().Be(0);
+                pool.NumberOfInUseConnections.Should().Be(2);
+
+                var timmer = new Stopwatch();
```
lutovich: Typo :) `timmer` -> `timer`

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
lutovich: I initially thought about connection acquisition timeout as being a per-attempt thingy. However it probably makes more sense to have it per the whole `Pool#acquire()` call. It is a user-facing setting and users can only reason about total time it takes to acquire a connection from the pool.

Neo4j.Driver/Neo4j.Driver/V1/Config.cs
```diff
@@ -77,6 +77,11 @@ public class Config
         [System.Obsolete("Do not set idle connection size to infinite.")]
         public const int InfiniteMaxIdleSessionPoolSize = -1;
 
+        /// <summary>
+        /// This const sets the number of connections in connection pool <see cref="MaxConnectionPoolSize"/> to be inifinite,
+        /// </summary>
+        public const int InfiniteMaxConnectionPoolSize = -1;
```
lutovich: Maybe we can just get away with documenting that `-1` means infinite? Otherwise it is strange that we have constant only for this setting and not for `ConnectionAcquisitionTimeout` or `MaxConnectionPoolSize`

lutovich (commented):
@zhenlineo, @ali-ince changes generally look good. I agree that connection acquisition timeout should be for the whole acquisition attempt. This can be addressed in a separate PR.

ali-ince:
Changed some naming about Config properties and constants. Also removed the recommendation of setting MaxConnectionPoolSize on async scenerios (at least for now) based on the previous benchmark results.

PR #229, opened 2017-08-25
Added tests for ipv6 address parsing for routing procedure responses

zhenlineo:
The parsing of ipv6 logic was not changed for routing procedure responses as it originally could accept ipv6 addresses.

While this parsing logic depends on the assumption that the addresses in procedure replies should be in the format of `address:port`. The `port` must be included.


zhenlineo:
@martinfurmanski Could you confirm if I could assume that the procedure reply will always includes `port` part?

martinfurmanski:
@zhenlineo @ali-ince Sorry for the late reply. Yes, the port will always be included!

It will be `[address]:port` or `address:port`.

ali-ince:
@martinfurmanski if there's a possibility of address:port for IPv6 addresses, then we need to make a couple of more changes.

martinfurmanski:
@ali-ince IPv6 will always be `[address]:port`.

PR #230, opened 2017-08-29
Tidy up some public facing API

ali-ince:
Removed ISessionAsync, ITransactionAsync & IStatementRunnerAsync interfaces and merged their contents with ISession, ITransaction & IStatementRunner.


lutovich (approved):
@ali-ince LGTM

PR #231, opened 2017-09-05
1.5 change test project targets

ali-ince:


Neo4j.Driver/common.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+	<PropertyGroup>
+		<XunitVersion>2.3.0-beta2-*</XunitVersion>
```
zhenlineo, line 3: Do we need to use beta version? Seems like they are approaching GA soon. https://www.nuget.org/packages/xunit/

ali-ince: In order to use the new cross-platform testing infrastructure, this is the package we've to use. However, since this is an improvement we can wait for its release and leave this PR open.

d0pare:
@ali-ince  [xunit 2.3.1](https://www.nuget.org/packages/xunit/2.3.1) was released

zhenlineo:
Pls move this branch to target latest driver version. The changes look very helpful.

PR #233, opened 2017-09-13
Improve how ConnectionAcquisitionTimeout is checked

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -174,118 +169,160 @@ public IConnection Acquire(AccessMode mode)
...
                     }
-                    ThrowObjectDisposedException();
-                }
 
-                return connection;
-            });
+                    return connection;
+                });
+            }
+            catch (OperationCanceledException ex)
```
zhenlineo: I am thinking that this `try-catch` should be inside `return TryExecute`. TryExecute is only used to log errors before throwing the error to user. I would think this `OperationCanceledException` is less interesting to an end user than `ClientException`

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -174,118 +169,160 @@ public IConnection Acquire(AccessMode mode)
...
-                            break;
+                            do
+                            {
+                                if (!IsConnectionPoolFull())
+                                {
+                                    connection = CreateNewPooledConnection();
+                                    break;
+                                }
+
+                                if (_availableConnections.TryTake(out connection, 0, cancellationToken))
```
zhenlineo: Why removed `ConnAcquisitionQueuingTimeout`? We initially want this blocking queue is to make thread wait a bit on this `TryTake` method to introduce some queuing/waiting inside async. I am a bit worried if we return directly without any sleep, will the thread span here a lot?

zhenlineo (commented):
Totally agree with the CancelationToken conn acquisition part, but I am a bit concerned to totally remove `ConnAcquisitionQueuingTimeout`.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
ali-ince: I've put a small time lapse back into the loop (not a magnitude of seconds but milliseconds) but remove those timeouts and cancellationtoken checking logic from the first TryTake - making the fast path do not bother with those.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -174,118 +171,161 @@ public IConnection Acquire(AccessMode mode)
...
+
+                        if (!_connectionValidator.IsValid(connection))
+                        {
+                            DestoryConnection(connection);
                         }
-                        if (_availableConnections.TryTake(out connection, ConnAcquisitionQueuingTimeout))
+                        else
                         {
                             break;
+                            ;
```
lutovich: Is this `;` intentional?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -174,118 +171,161 @@ public IConnection Acquire(AccessMode mode)
...
+                            ThrowObjectDisposedException();
+                        }
+
+                        if (!_availableConnections.TryTake(out connection))
+                        {
+                            do
+                            {
+                                if (!IsConnectionPoolFull())
+                                {
+                                    connection = CreateNewPooledConnection();
```
lutovich, line 203: Unrelated to this PR: given a pool with `maxConnectionPoolSize - 1` acquired and `0` idle connections, can two threads that call `Acquire()` concurrently make pool expand to `maxConnectionPoolSize + 1` active connections? They will both try to create concurrently and attempts are probably not synchronized.

lutovich (commented):
@ali-ince looks good, two small comments

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
ali-ince: No, indeed :)

ali-ince:
I think I've addressed the synchronization issue you've pointed out.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Except for correctness, I am not fan to strictly not exceed maxConnPoolSize. As it not only slows the code down also bring more complete logic.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -151,6 +149,17 @@ public IPooledConnection CreateNewPooledConnection()
         private PooledConnection NewPooledConnection()
         {
             Interlocked.Increment(ref _poolSize);
+            if (_maxPoolSize != Config.Infinite && _poolSize > _maxPoolSize)
+            {
+                lock (_syncObject)
+                {
+                    if (_poolSize > _maxPoolSize)
+                    {
+                        throw new InvalidOperationException();
```
lutovich: Should we also decrement `_poolSize` here?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -151,6 +149,17 @@ public IPooledConnection CreateNewPooledConnection()
         private PooledConnection NewPooledConnection()
         {
             Interlocked.Increment(ref _poolSize);
+            if (_maxPoolSize != Config.Infinite && _poolSize > _maxPoolSize)
```
lutovich: If we have two threads and `_poolSize == _maxPoolSize  - 1` they both will increment `_poolSize` and throw, right? I think one of them should succeed. We could probably use result of `Interlocked.Increment` to figure out which one should fail.

lutovich (commented):
@ali-ince two more comments about synchronization around `maxPoolSize`

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Suggesting to have the fix in another PR. Right now the scope of this PR is good enough and the problem is not critical to block 1.5.

PR #235, opened 2017-09-15
Fix for Keys availability upon Session.RunAsync and Transaction.RunAsync

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/Transaction.cs
```diff
@@ -222,6 +223,10 @@ public override Task<IStatementResultCursor> RunAsync(Statement statement)
                     _connection.Server);
                 _connection.Run(statement.Text, statement.Parameters, resultBuilder);
                 await _connection.SendAsync().ConfigureAwait(false);
+
+                // Wait for SUCCESS/FAILURE message from the server
+                await _connection.ReceiveOneAsync().ConfigureAwait(false);
```
zhenlineo, line 228: I thought for transaction run, if we did not provide bookmark, at line 225 we would have queued messages:

```
Run Begin
Pull_All
Run transaction XX
Pull_All
```

Then if we send and just wait for the first reply, then we only got reply for `Run Begin`?

I am now wondering why tests would pass...

zhenlineo (changes requested):
Can we also `Fix` the sync Result to be the same as java where we will wait for the first Success if the user calls result.Keys?

zhenlineo (approved):

Neo4j.Driver/Neo4j.Driver/Internal/Transaction.cs
zhenlineo: I tested locally, the `receiveOneAsync` will only wait for `Success` for the first`Run Begin` message. While as we batching messages and handle them together if them fit in buffer, when `Success` for `Run Begin` returns, the others messages also arrived and processed in time.

PR #236, opened 2017-09-19
added IDriver.Close and IDriver.CloseAsync methods.

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -189,7 +203,7 @@ private IPooledConnection Acquire(CancellationToken cancellationToken)
                 {
                     while (true)
                     {
-                        if (_disposeCalled)
+                        if (_closedMarker > 0)
```
lutovich: Maybe extract `_closedMarker > 0` in a method like `IsClosed()`?
It is used in many places and feels a little bit cryptic.

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketClient.cs
```diff
@@ -40,6 +40,9 @@ private static class ProtocolVersion
         private IBoltWriter _writer;
         private readonly BufferSettings _bufferSettings;
 
+        private int _closedMarker = 0;
+        private bool _isOpen = false;
```
lutovich: Could we remove this flag and only keep `closedMarker`?
I think `closedMarker` can be initialized with `-1` and we can set it to `0` where we currently do `_isOpen = true`.

Neo4j.Driver/Neo4j.Driver/Internal/Driver.cs
```diff
@@ -100,19 +78,61 @@ public ISession Session(IEnumerable<string> bookmarks)
 
         private ISession Session(AccessMode defaultMode, Bookmark bookmark)
         {
-            if (_disposeCalled)
+            if (_closedMarker > 0)
```
lutovich: Maybe also extract a method

lutovich (commented):
@ali-ince LGTM. Made couple small comments. Maybe worth extracting all `_closedMarker > 0` in `IsClosed` methods to make them more readable.

zhenlineo:
LG. will merge once the build is back to green.

PR #241, opened 2017-09-21
Changed the impl for waiting for keys to arrive

zhenlineo:
For sync, the key will be received when `result.Keys` method is called. The method will block there until the key arrives via the reply of `Run` message

For async, the key will be received when a result cursor is built. While returning cursor after `Run` method, we will wait there for the reply of `Run` message

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
@@ -80,10 +80,7 @@ namespace Neo4j.Driver.Internal
 
                 await _connection.SendAsync().ConfigureAwait(false);
 
-                // Wait for SUCCESS/FAILURE message from the server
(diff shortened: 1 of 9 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -82,10 +82,7 @@ public override Task<IStatementResultCursor> RunAsync(Statement statement)
 
                 await _connection.SendAsync().ConfigureAwait(false);
 
-                // Wait for SUCCESS/FAILURE message from the server
-                await _connection.ReceiveOneAsync().ConfigureAwait(false);
-
-                return resultBuilder.PreBuild();
+                return await resultBuilder.PreBuildAsync();
```
ali-ince: ConfigureAwait is missing here.

zhenlineo:
Fixed quickly now :P

ali-ince (approved):

PR #242, opened 2017-09-25
improve pool size checks

ali-ince:


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -150,7 +151,12 @@ public IPooledConnection CreateNewPooledConnection()
 
         private PooledConnection NewPooledConnection()
         {
-            Interlocked.Increment(ref _poolSize);
+            var currentPoolSize = Interlocked.Increment(ref _poolSize);
+            if (_maxPoolSize != Config.Infinite && currentPoolSize > _maxPoolSize)
+            {
+                throw new InvalidOperationException();
```
zhenlineo: Maybe return null? As exception is still more expensive than a boolean check. This pooling code I would prefer it to be as fast as possible.

zhenlineo (changes requested):
LG, I got a small improvement suggestion

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
ali-ince: I intended the same at first, but did not continue as it would result in a less straight-forward code (with some if and else statements) for a mostly rare use case (with max pool size set and already reached the highest watermark with several competing threads).

lutovich: Can out parameters be used to remove this exception?
Smth like `bool NewPooledConnection(out PooledConnection connection)` which returns `true` when connection was created and `false` otherwise.

zhenlineo (approved):

Issue #244, opened 2017-09-25
What is the idea of how to read values values out of PackStreamStruct?


flq:
Consider the following  statement:

```
MATCH (p:Person)-[r:OWNS]->(n) 
WITH p, collect(n) as products
WHERE id(p) = {id} RETURN p, products
```

Here the _products_ key will point to a `List<object>` whose elements are `Node`s, fair enough.

Now, suppose I also need relationship info and do something like this:

```
MATCH (p:Person)-[r:OWNS]->(n) 
WITH p, collect({ rel: r, product: n }) as products
WHERE id(p) = {id} RETURN p, products
```

What I now get is

```
((Dictionary<string,object>)((List<object>)result[0]["products"])[0]).Values
Count = 2
    [0]: {Neo4j.Driver.Internal.IO.PackStreamStruct}
    [1]: {Neo4j.Driver.Internal.IO.PackStreamStruct}

```
And here I cannot sensibly access the values of **rel** and **product** anymore because the value is represented with an internal object.

If you tell me that this is a fairly stable API I can drill holes into it to extract the values, but maybe you tell me that you will actually make those items public?

Looking forward to any pointers

zhenlineo:
`PackStreamStruct` should stay internal. We probably exposed the wrong type here. Working on a fix now.

Thanks so much for reporting to us!
Zhen

flq:
Thank you Zhen,
as an aside I was looking at whether I can put myself in the position to contribute since my current "pet" project involves knowing the .NET Neo Driver pretty well - running the integration & TCK(?) tests, it appears that you require access to some AWS resources...

`Required environment variable is not defined: AWS_ACCESS_KEY_ID`

However, the "Tests" project runs fine.

zhenlineo:
@flq Yeah, IT and TCK tests requires aws to download a server to run all tests unless you have already downloaded a server by running tests before. You could probably run most of IT and TCK tests if you have a community server in the path `github-parent-folder/Target/neo4jhome`, with password `neo4j` (See more in this [class](https://github.com/neo4j/neo4j-dotnet-driver/blob/230d8cefabc2d8c5c76db42c50049ac843c0d6c0/Neo4j.Driver/Neo4j.Driver.IntegrationTests/Internals/StandAlone/ExternalBoltkitInstaller.cs)). But any ITs that has `RoutingDriver` or `Cluster` in its name still requires a enterprise server.

[...]

PR #245, opened 2017-09-25
Fixed internal class leakage

ali-ince:
1. Removed PackStreamStruct class,
  2. Introduced struct handlers which are PackStream integrated by default.


zhenlineo:
Fix for https://github.com/neo4j/neo4j-dotnet-driver/issues/244

PR #256, opened 2017-11-13
Adds CancellationToken in run method

zhenlineo:
The cancellation token applies both to Run and streaming result, a.k.a. The cancellation token will be passed from `Session/Tx.RunAsync` to `ResultStatement.NextAsync`. A user has to finish consuming the result before the cancellationSource is cancelled.

When a cancellationToken passed to a session/tx.run method is cancelled, a `RESET` message will be sent to the server to cancel the statement running/streaming. The cancellationToken cancel event will NOT stop or interrupt any existing network read operation on the client side. It just signal the server to cancel the statement execution. It is the server who will stop to run or stream the result back. If the server is not interrupted by `RESET` (for some reasons a.k.a. a bug?), then the client will remain blocked.

This PR changed the place where to ack failure

**Mute ack failure:**
We should disallow client to send `ACKFAIL` if the client already sent `RESET`. This is because `RESET` will clean all failure state on server side. Then if we still send `ACKFAIL` after `RESET`, the server will complain that there is no error to ack. Therefore we need to mute `ACKFAIL` from a `RESET` is sent till the reply of `RESET` is received.

**Ack failure when the failure comes:**
After we changed to pull as much as messages from `ChunkedInput`, we might receive and handle more than one messages in `SocketConnection.ReceiveOne` method. While we ack failure for any server error only at the end of this method. It could be too late to let the `AckFailure` method know whether it can ack or not. 

Here is an example: The client sends `RUN`, `PULLALL` and `RESET` to the server, and then the server replies `FAIL`, `IGNORED`, and `SUCC` in one chunk. If we ack failure after we read all messages in this chunk, then we already passed the time slot when a `RESET` is sent till the reply of `RESET` arrives. We would think we could `ACKFAIL` now. However the truth is we should not in this case. This problem would go away if we ack failure at the moment the failure occurs.

**Changes**
AckFail is a protocol. Therefore when we receive an error, we will try to ack fail. Then the connection will help us to test if it can ack or not and enqueue the messages if it can ack.

PooledConnection will no longer handle ack failure, instead, it only cares about if the connection has any unrecoverable error or not. If it has, then this connection will be closed. If it does not, then it will be reused.

This PR also fixed a bug in `DelegatedConn` where an error from chained tasks passed to next call as an aggregated error rather than its base error.


zhenlineo:
Write a test to confirm tx is rolledback properly after cancelled

PR #257, opened 2017-11-15
Fix purge connections based on `AggregateException` rather the original error

zhenlineo:
Fix the bug where an error was handled as `AggregateException` rather than the original error in Connections

Clean up some unreached code


ali-ince (approved):
LGTM!

PR #260, opened 2017-11-21
Do not purge inUse connections

zhenlineo:
This PR consists of two parts:
1) When updating routing table, the pool of the servers that are removed from routing table should not be closed but be deactivated.
2) When a connection error happens towards a server, the pool of the server should not be purged but deactivated. 

TODOs (Not in this PR but comes later):
* While working on 2) we found that the `DelegatedConnection.OnError` should also provides a Async impl so that the connections could be closed synchronically. This will be left to some future work. 
* Chain async tasks if possible

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -1362,7 +1362,7 @@ namespace Neo4j.Driver.Tests
             }
         }
 
-        public class DeactiviateMethod
+        public class DeactivateMethod
         {
             private static List<Mock<IPooledConnection>> FillIdleConnections(
                 BlockingCollection<IPooledConnection> idleConnections, int count)
@@ -1443,7 +1443,7 @@ namespace Neo4j.Driver.Tests
             // concurrent test
             // concurrently close and deactive
             [Fact]
-            public void DeactiviateAndThenCloseShouldCloseAllConnections()
+            public void DeactivateAndThenCloseShouldCloseAllConnections()
             {
                 var idleConnections = new BlockingCollection<IPooledConnection>();
                 var idleMocks = FillIdleConnections(idleConnections, 5);
@@ -1486,7 +1486,7 @@ namespace Neo4j.Driver.Tests
                 pool.NumberOfInUseConnections.Should().Be(0);
 
                 // This is to simulate Acquire called first,
-                // but before Acquire put a new conn into inUseConn, Deactiviate get called.
+                // but before Acquire put a new conn into inUseConn, Deactivate get called.
                 openConnMock.Setup(x => x.IsOpen).Returns(true)
                     .Callback(() => pool.Deactivate());
                 idleConnections.Add(openConnMock.Object);
@@ -1511,7 +1511,7 @@ namespace Neo4j.Driver.Tests
                 pool.NumberOfInUseConnections.Should().Be(0);
 
                 // This is to simulate Acquire called first,
-                // but before Acquire put a new conn into inUseConn, Deactiviate get called.
+                // but before Acquire put a new conn into inUseConn, Deactivate get called.
                 // However here, this connection is not healthy and will be destoried directly
                 closedConnMock.Setup(x => x.IsOpen).Returns(false)
                     .Callback(() => pool.Deactivate());
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -546,7 +546,8 @@ namespace Neo4j.Driver.Tests
                     pool.Acquire();
                 }
 
-                var stopWatch = new Stopwatch(); stopWatch.Start();
+                var stopWatch = new Stopwatch();
+                stopWatch.Start();
 
                 var exception = await Record.ExceptionAsync(() => pool.AcquireAsync(AccessMode.Read));
 
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -1182,17 +1182,17 @@ namespace Neo4j.Driver.Tests
 
         public class PoolState
         {
-            // open
+            // active
             [Fact]
-            public void FromOpenViaAcquireToOpen()
+            public void FromActiveViaAcquireToActive()
             {
                 var pool = NewConnectionPool();
                 pool.Acquire();
-                pool.Status.Should().Be(PoolStatus.Open);
+                pool.Status.Should().Be(PoolStatus.Active);
             }
 
             [Fact]
-            public void FromOpenViaReleaseToOpen()
+            public void FromActiveViaReleaseToActive()
             {
                 var idleQueue = new BlockingCollection<IPooledConnection>();
                 var inUseConnections = new ConcurrentSet<IPooledConnection>();
@@ -1204,11 +1204,11 @@ namespace Neo4j.Driver.Tests
 
                 idleQueue.Count.Should().Be(1);
                 inUseConnections.Count.Should().Be(0);
-                pool.Status.Should().Be(PoolStatus.Open);
+                pool.Status.Should().Be(PoolStatus.Active);
             }
 
             [Fact]
-            public void FromOpenViaDisposeToClosed()
+            public void FromActiveViaDisposeToClosed()
             {
                 var pool = NewConnectionPool();
                 pool.Dispose();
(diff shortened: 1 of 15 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnection.cs
```diff
@@ -47,7 +48,35 @@ public override void OnError(Exception error)
...
             }
-            else if (error.IsClusterError())
+            else
+            {
+                HandleClusterError(error);
+            }
+            throw error;
+        }
+
+        public async Task OnErrorAsync(Exception error)
```
zhenlineo, line 58: @ali-ince This is the method that need to be plugged in `DelegatedConnection`

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -428,11 +441,11 @@ public void Release(IPooledConnection connection)
                     }
                     else
                     {
-                        _availableConnections.Add(connection);
+                        _idleConnections.Add(connection);
                     }
 
                     // Just dequeue any one connection and close it will ensure that all connections in the pool will finally be closed
-                    if (IsClosed && _availableConnections.TryTake(out connection))
+                    if (IsZombieOrClosed && _idleConnections.TryTake(out connection))
```
ali-ince: Shall we consider adding ```IsZombieOrClosed || ``` check also to [L438](#L438)

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -468,11 +481,11 @@ public Task ReleaseAsync(IPooledConnection connection)
...
```
ali-ince: Shall we consider adding ```IsZombieOrClosed || ``` check also to [L478](#L478)?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnectionPool.cs
```diff
@@ -116,26 +115,62 @@ public void Add(IEnumerable<Uri> servers)
         {
             foreach (var uri in servers)
             {
-                Add(uri);
+                if (_pools.ContainsKey(uri))
```
ali-ince: It may be better and safer to make use of ```_pools.AddOrUpdate(uri, CreateNewConnectionPool, ActivateConnectionPool)``` where `ActivateConnectionPool` is a new method that just activates and returns the activated pool instance.

ali-ince: And also we can change `GetOrAdd` to `TryAdd` at [L104](#L104).

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -263,20 +275,20 @@ private IConnection CreateClusterConnection(Uri uri, AccessMode mode)
...
                     $"Routing table {_routingTableManager.RoutingTable} contains a server {uri} " +
-                    $"that is not known to cluster connection pool {_clusterConnectionPool}."));
+                    $"that is not known to cluster connection pool {_clusterConnectionPool}.")).ConfigureAwait(false);
             }
             catch (ServiceUnavailableException e)
             {
-                OnConnectionError(uri, e);
+                await OnConnectionErrorAsync(uri, e).ConfigureAwait(false);
             }
             return null;
```
ali-ince, line 291: We probably did not notice the problem here before, but we should be returning `Task.FromResult(null)` instead of `null`.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnection.cs
ali-ince: I think adding `DelegatedConnection.OnErrorAsync` makes sense. The only actual place where async job is done seems to be `ClusterConnection` though. This also brings another option of introducing `IAsyncErrorHandler` interface with `OnErrorAsync`  and selectively calling it from within `DelegatedConnection.TaskWithErrorHandling`?

ali-ince (commented):
Looks good to me, with a couple of small comments.

ali-ince:
I really do not know whether we should do or it would worth it but @lutovich introduced a fallback mechanism for Java Driver at [LoadBalancer](https://github.com/neo4j/neo4j-java-driver/blob/1.4/driver/src/main/java/org/neo4j/driver/internal/cluster/LoadBalancer.java#L36). What do you think?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
zhenlineo: Seems we do not need to cast as we have `async` keywords in the method declaration. And it will cast it to a Task<null> automatically.

ali-ince: Of course!

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
```diff
@@ -1133,5 +1180,413 @@ public void ShoulReportPoolSizeCorrectOnConcurrentRequests()
...
+            // zombie
+            [Fact]
+            public void FromZombieViaAcquireThrowsError()
+            {
+                var pool = NewConnectionPool();
+                pool.Status = PoolStatus.Zombie;
+
+                var exception = Record.Exception(()=>pool.Acquire());
+
+                exception.Should().BeOfType<ObjectDisposedException>();
```
lutovich: Maybe worth throwing some other exception when acquiring from zombie pool?
I thought `ObjectDisposedException` should only be thrown after `Dispose()` invocation.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -24,43 +24,54 @@
...
     internal class ConnectionPool : LoggerBase, IConnectionPool
     {
         private const int SpinningWaitInterval = 500;
 
         private readonly Uri _uri;
 
-        private int _closedMarker = 0;
+        private int _poolStatus = Open;
+        private bool IsClosed => _poolStatus == Closed;
+        private bool IsZombieOrClosed => _poolStatus != Open;
```
lutovich: Field `_poolStatus` is modified using `Interlocked` but is read using regular field accesses, is it ok?

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnectionPool.cs
```diff
@@ -70,15 +71,9 @@ internal class ClusterConnectionPool : LoggerBase, IClusterConnectionPool
...
-        private IConnectionPool CreateNewConnectionPool(Uri uri)
-        {
-            return _fakePool ?? new ConnectionPool(uri, _connectionSettings, _poolSettings, _bufferSettings, Logger);
-        }
-
         public IConnection Acquire(Uri uri)
         {
-            IConnectionPool pool;
-            if (!_pools.TryGetValue(uri, out pool))
+            if (!_pools.TryGetValue(uri, out var pool))
```
lutovich, line 76: Wow!

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnectionPool.cs
```diff
@@ -99,72 +93,113 @@ public Task<IConnection> AcquireAsync(Uri uri)
...
+                    }
                 }
             }
-            foreach (var uri in servers)
+        }
+
+        public async Task UpdateAsync(IEnumerable<Uri> added, IEnumerable<Uri> removed)
+        {
+            await AddAsync(added).ConfigureAwait(false);
+            // TODO chain this part and use task.waitAll
```
lutovich, line 145: Is this TODO still relevant?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -24,43 +24,54 @@
...
+using static Neo4j.Driver.Internal.PoolStatus;
+using static Neo4j.Driver.Internal.Throw.ObjectDisposedException;
 
 namespace Neo4j.Driver.Internal
 {
+    internal static class PoolStatus
+    {
+        public const int Open = 0;
+        public const int Closed = 1;
+        public const int Zombie = 2;
```
lutovich: Maybe rename "open" to "active" and "zombie" to "inactive"?

lutovich (commented):
@zhenlineo changes look good to me. Made couple small comments. Is it possible to add cluster IT for this new behaviour?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: Good point!

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: How about `alive`, `zombie` and `totallyDead`? :P I will renaming as you suggested.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterConnectionPool.cs
zhenlineo: This is the second TODO. Will do in another PR.

Issue #250, opened 2017-10-18
Big requests do not complete


juicelink:
On my solution if I run:
`session.Run("match(n) return n limit 5000").ToList()`
It works

if I run:
`session.Run("match(n) return n limit 10000").ToList()`
It never completes

It works with 1.5.0-alpha01 (result comes very quickly) but not with the upper versions (final 1.5.0 included).

(test with dotnet core 2 in xunit unit test)

ali-ince:
Hi @juicelink,

Just as a quick check, I've executed [this code](https://gist.github.com/ali-ince/cf554c9e1eeba02252236dd59ab3a4a8) successfully on both .NET Framework 4.6.2 and .NET Core 2.0 as a console app.

Can you give us more information about the environment (OS, neo4j server version, whether this code connects to a remote database, etc.) and any specific config options set on Neo4j.Driver? 

Ali

juicelink:
Hi @ali-ince,

Neo4j server: 3.2.1
Server os: Windows server 2012 R2
Client OS: Windows 7 Enterprise SP1
Yes the code connects to a remote database

here's my driver initialization:
`GraphDatabase.Driver(
    "bolt://host:port/",
    AuthTokens.Basic("user", "pwd"),
    new Config { EncryptionLevel = EncryptionLevel.None, ConnectionTimeout = ConnectionTimeout(3600)})`

If I use FetchAsync, it returns the first 3000 - 4000 records then it blocks;

AviAvni:
Hi we have this problem too from what we debug it's look like this method is the problem
https://github.com/neo4j/neo4j-dotnet-driver/blob/1.5/Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs#L113
we alyays get in the trace that the 'read' variable is 0

AviAvni:
sorry it's in https://github.com/neo4j/neo4j-dotnet-driver/blob/1.5/Neo4j.Driver/Neo4j.Driver/Internal/IO/ChunkReader.cs#L166
when ReadAsync is task result is 0 it dosn't stop read from the stream

@zhenlineo please look at this it's blocking us from using 1.5 in production and it makes the cpu to 100%

AviAvni:
wrap the code with await session.ReadTransactionAsync(async transaction
solve the issue

ali-ince:
Hi,

[...]

PR #266, opened 2017-12-04
Compact buffer when writable bytes drop below a set low watermark

ali-ince:
Internal buffer used by `ChunkedReader` was not being reset  (its recorded positions) under edge cases, where `Read` operation on `NetworkStream` was given a `count` of `0` which was conceived as an `Unexpected end of stream`.

This PR addresses this problem by checking the number of writable bytes left in the buffer against a pre-defined constant value of `1024` and when drops under this value, the buffer positions are reset and leftover bytes are copied to the beginning of the buffer.

Related with the discussion on [Introduce network level write throttling #10433](https://github.com/neo4j/neo4j/pull/10433) and might be related to #250.


AviAvni:
@ali-ince I downloaded this code and compiled it and it's solve the problem I reported

PR #270, opened 2018-01-08
Refactoring for multiple Bolt Protocol support

zhenlineo:
Extracted a factory to create protocol based on the handshake result.
Simplified some tests


lutovich (approved):
Changes look good to me.

ali-ince (approved):
👍

PR #271, opened 2018-01-12
Improved connection pool statistics

zhenlineo:
And it is called Metrics, as we only aim to providing raw measurements.

A new dependency HdrHistogram is introduced to better support histogram used in our metrics.
As a dependency requirement, we upgrade to .net core 1.6.1.

There are two types of metrics in this driver: a connection metrics and a connection pool metrics.
The metrics are collected via listeners added in connections and connection pools. The listener interface is not directly exposed to end user as we do not want to have surprises from user's code. Instead the metrics are returned via `driver.GetDriverMetrics()`.

Example of usage:

```
var config = new Config {DriverMetricsEnabled = true};
var driver = GraphDatabase.Driver(uri, auth, config);
var metrics = driver.GetDriverMetrics(); // an exception is thrown if metrics is not enabled.
```

Then we could access metrics such as the code shown bellow:
```
var inUseConnCount = metrics.PoolMetrics[Name].InUse
var acquisitionDelayMedian = metrics.PoolMetrics[Name].AcquisitionTimeHistogram. GetValueAtPercentile(50);

var connDelayHistogram = metrics.ConnectionMetrics[Name].ConnectionTimeHistogram
var curMaxConnDelay = connDelayHistogram.Max;
connDelayHistogram.Reset();
```


zhenlineo:
@ali-ince See #272 instead. Fixed a bug introduced in this PR.

PR #272, opened 2018-01-26
Added stress tests for routing driver

zhenlineo:
The main purpose of this PR is to add a stress test for routing driver to ensure that the load balancer and pool could handle random kill of connections (random down of servers)

As a result,
Refactored to use factory pattern to create connections and pools to make it easier to inject test impl of connection/pool.

Based on #271 
Fixed a bug in `Metrics` where we cannot add deactivated pool metrics back due to dictionary key collision.


Neo4j.Driver/Neo4j.Driver.IntegrationTests/RoutingDriver/RoutingDriverStressIT.cs
```diff
@@ -0,0 +1,213 @@
...
+            var bufferSettings = new BufferSettings(config);
+            var connectionFactory = new MonitoredPoolledConnectionFactory(
+                new PooledConnectionFactory(connectionSettings, bufferSettings, config.Logger));
+
+            _driver = (Internal.Driver) GraphDatabase.CreateDriver(new Uri(RoutingServer), config, connectionFactory);
+            _connections = connectionFactory.Connections;
+            _metrics = _driver.GetDriverMetrics();
+        }
+
+        private Task ConnectionTerminator()
```
ali-ince, line 145: I could not understand the reason of having this connection killing logic, shouldn't `MaxIdleConnectionPoolSize` config 
 be an option?

Neo4j.Driver/Neo4j.Driver.IntegrationTests/Shared/SoakRunWorkItem.cs
```diff
@@ -124,21 +97,113 @@ public Task Run()
...
+                {
+                    var currentIteration = Interlocked.Increment(ref _counter);
+                    var query = queries[currentIteration % queries.Length];
+                    var accessMode = accessModes[currentIteration % accessModes.Length];
+
+                    using (var session = _driver.Session())
+                    {
+                        try
+                        {
+                            void RunTx(ITransaction tx)
```
ali-ince: This can be re-written for better readability like

```csharp
Action<ITransaction> runTx = (tx) => tx.Run(query).Consume();
```

Neo4j.Driver/Neo4j.Driver.IntegrationTests/Shared/SoakRunWorkItem.cs
```diff
@@ -124,21 +97,113 @@ public Task Run()
...
+            for (var i = 0; i < repeat; i++)
+            {
+                var currentIteration = Interlocked.Increment(ref _counter);
+                var query = queries[currentIteration % queries.Length];
+                var accessMode = accessModes[currentIteration % accessModes.Length];
+
+                var session = _driver.Session(accessMode);
+                try
+                {
+                    async Task RunTxAsync(ITransaction tx)
```
ali-ince: Similarly, could be written as

```csharp
Func<ITransaction, Task> runTxAsync = async (txc) => 
  {
    var result = await txc.RunAsync(query);
    await result.ConsumeAsync();
  };
```
or even

```csharp
Func<ITransaction, Task> runTxAsync = async (txc) =>
    await (await txc.RunAsync(query)).ConsumeAsync();
```

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -34,6 +32,25 @@ internal static class PoolStatus
         public const int Active = 0;
         public const int Closed = 1;
         public const int Inactive = 2;
+
+        private static readonly IDictionary<int, string> StatusCodeToNameMappingTable = new Dictionary<int, string>
```
ali-ince: What do you think about changing `PoolStatus` to a [type safe enum pattern](https://www.infoworld.com/article/3198453/application-development/how-to-implement-a-type-safe-enum-pattern-in-c.html)?

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -196,9 +195,9 @@ private void DestroyConnection(IPooledConnection conn)
                 return;
             }
 
-            _statistics?.IncrementConnectionToClose();
+            _poolMetricsListener?.BeforeConnectionClosed();
             conn.Destroy();
-            _statistics?.IncrementConnectionClosed();
+            _poolMetricsListener?.AfterConnectionClosed();
```
ali-ince: Do you think it's a good idea that we have this and the other similar occurrences in a `finally` clause?

Neo4j.Driver/Neo4j.Driver/Internal/IPooledConnection.cs
```diff
@@ -28,12 +28,15 @@ internal interface IPooledConnection : IConnection
...
 
         /// <summary>
-        /// Try to reset the connection to a clean state to prepare it for a new session.
+        /// Try to reset the connection to a clean state.
         /// </summary>
         void ClearConnection();
-
         Task ClearConnectionAsync();
 
+        void OnRequire();
```
ali-ince: Maybe `OnAcquire`?

Neo4j.Driver/Neo4j.Driver/Internal/Metrics/ConnectionPoolMetrics.cs
```diff
@@ -0,0 +1,113 @@
...
+        public long Closed => _closed;
+        public long FailedToCreate => _failedToCreate;
+
+        public int ToCreate => _toCreate;
+        public int ToClose => _toClose;
+
+        public int InUse => _pool?.NumberOfInUseConnections ?? 0;
+        public int Idle => _pool?.NumberOfIdleConnections ?? 0;
+
+        private readonly ConcurrentSet<IListenerEvent> _acquisitionDelayTimers = new ConcurrentSet<IListenerEvent>();
```
ali-ince: Do we use this field? I see we're adding at line 93 and removing at line 100, but don't check for existence anywhere.

Neo4j.Driver/Neo4j.Driver/Internal/Metrics/DriverMetrics.cs
```diff
@@ -0,0 +1,73 @@
...
+            ConnectionMetrics = new ConcurrentDictionary<string, IConnectionMetrics>();
+        }
+
+        public ConnectionPoolMetrics AddPoolMetrics(Uri poolUri, IConnectionPool pool)
+        {
+            var acquisitionTimeout = _config.ConnectionAcquisitionTimeout;
+            var poolMetrics = new ConnectionPoolMetrics(poolUri, pool, acquisitionTimeout);
+            var key = poolMetrics.UniqueName;
+
+            if (PoolMetrics.ContainsKey(key))
```
ali-ince: This and the next similar block could be written as follows, only if we had a private field of type `ConcurrentDictionary<string, IConnectionPoolMetrics>`.

```csharp
PoolMetrics.AddOrUpdate(key, poolMetrics, (oldKey, oldValue) => poolMetrics);
```

Neo4j.Driver/Neo4j.Driver/Internal/Metrics/IConnectionListener.cs
```diff
@@ -0,0 +1,55 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Diagnostics;
+
+namespace Neo4j.Driver.Internal.Metrics
+{
+    internal interface IConnectionListener
```
ali-ince, line 23: Maybe add `BeforeDisconnect` and `AfterDisconnect` to measure hard disconnects, does it makes sense?

Neo4j.Driver/Neo4j.Driver.IntegrationTests/RoutingDriver/RoutingDriverStressIT.cs
```diff
@@ -0,0 +1,213 @@
...
+                }
+                catch (TaskCanceledException)
+                {
+                    // we are fine with cancelled sleep
+                }
+            }
+        }
+
+
+        private class MonitoredPoolledConnectionFactory : IPooledConnectionFactory
```
ali-ince: Maybe `MonitoredPooledConnectionFactory`?

ali-ince (commented):
Changes look good for me. I made a couple of comments, some of them are misspelled words.

I think we should also consider some public accessible metrics exposure, to be incorporated into consumer's own monitoring system (like graphite, etc.).

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
zhenlineo: The disadvantage of it is that we will have something like

```
sealed class PoolStatus {
public static PoolStatus Active = new PoolStatus(0, "Active");
public static PoolStatus Closed = new PoolStatus(1, "Closed");
public static PoolStatus Inactive = new PoolStatus(2, "Inactive");

public int Value;
public string Name;

private PoolStatus(int value, sting name)
{
  Value = value;
  Name = name;
}
```

Then in the code, when we need to refer to Active's value, then we need to use `PoolStatus.Value` which dose not looks good to read.
 For example it is a bit hard to read:

```Interlocked.Exchange(ref _poolStatus, Closed.Value) != Closed.Value```

File: Neo4j.Driver/Neo4j.Driver/Internal/Metrics/IConnectionListener.cs (hunk 9 above)
Comment (zhenlineo) [reply Neo4j.Driver/Neo4j.Driver/Internal/Metrics/IConnectionListener.cs:23]: Feels not so useful to have the timer on Connection.conn actually. So I would hold to do this in the first version and then leave it in next iteration if you also agree?

File: Neo4j.Driver/Neo4j.Driver.IntegrationTests/RoutingDriver/RoutingDriverStressIT.cs (hunk 1 above)
Comment (zhenlineo) [reply Neo4j.Driver/Neo4j.Driver.IntegrationTests/RoutingDriver/RoutingDriverStressIT.cs:145]: This killing log is to simulate there is something wrong on the connections between the server and client. The aim is to let the driver realize a connection error and remove the server from routing table and purge idle in connection pool, but the query should still success as we are running queries in retries.

The `MaxIdle` is not set on purpose so that it will default to the same size as `MaxConn`.

File: Neo4j.Driver/Neo4j.Driver/Internal/Metrics/IConnectionListener.cs (hunk 9 above)
Comment (ali-ince) [reply Neo4j.Driver/Neo4j.Driver/Internal/Metrics/IConnectionListener.cs:23]: Yeah, ok. The amount of time spent closing the connection really doesn't make sense. I somehow thought about the number of connections hard closed. Never mind, we can leave it for next iteration as you suggested.

File: Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs (hunk 4 above)
Comment (ali-ince) [reply Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs]: I think it is possible to use it like
```csharp
Interlocked.Exchange(ref _poolStatus, Closed) != Closed
```
If we have `_poolStatus` of type `PoolStatus`.

Not a big deal here, just thought it may look better.

PR #273, opened 2018-02-07
Fixed the doc build issue because of new dependency required.

zhenlineo:
Changed to build docs based on `net452` dll and xml files as this version is better supported by SHFB.

### DIFF
--- a/Neo4j.Driver.Docs.shfbproj
+++ b/Neo4j.Driver.Docs.shfbproj
@@ -7,19 +7,19 @@
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <SchemaVersion>2.0</SchemaVersion>
     <ProjectGuid>{e4682985-dabf-462a-997e-debdfb01ede9}</ProjectGuid>
-    <SHFBSchemaVersion>2015.6.5.0</SHFBSchemaVersion>
+    <SHFBSchemaVersion>2017.9.26.0</SHFBSchemaVersion>
     <!-- AssemblyName, Name, and RootNamespace are not used by SHFB but Visual Studio adds them anyway -->
     <AssemblyName>Documentation</AssemblyName>
     <RootNamespace>Documentation</RootNamespace>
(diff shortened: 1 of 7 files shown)


Neo4j.Driver.Docs.shfbproj
```diff
@@ -18,8 +18,8 @@
     <HtmlHelpName>Neo4j.Driver.Documentation</HtmlHelpName>
     <Language>en-US</Language>
     <DocumentationSources>
-      <DocumentationSource sourceFile="Neo4j.Driver\Neo4j.Driver\bin\Release\netstandard1.3\Neo4j.Driver.dll" />
-<DocumentationSource sourceFile="Neo4j.Driver\Neo4j.Driver\bin\Release\netstandard1.3\Neo4j.Driver.xml" /></DocumentationSources>
+      <DocumentationSource sourceFile="Neo4j.Driver\Neo4j.Driver\bin\Release\net452\Neo4j.Driver.dll" />
+<DocumentationSource sourceFile="Neo4j.Driver\Neo4j.Driver\bin\Release\net452\Neo4j.Driver.xml" /></DocumentationSources>
```
ali-ince, line 22: Both entries have source files pointing to `net452` artifacts, shall one of them be defined to point `netstandard1.3`?

ali-ince (approved):
LGTM!

I've added one comment of which I'm not sure.

Neo4j.Driver.Docs.shfbproj
zhenlineo: No, this is the file used by Sandcastle to create API docs for this driver. Previously we generate the docs based on `netstandard1.3` binaries (as we only support netcore). However Sandcastle does not support this new framework very well. It always has some docs entry missing here and there. Now as we also support `net452`, so I change to let Sandcastle to generating API docs from `net452` binaries instead.

So the dll file is the binary and xml contains some API docs. I do not think this change will cause any changes to what it looks like today.

PR #277, opened 2018-03-06
Improve metrics

zhenlineo:
Renamed some interface methods to be the same as Java driver, such as `DriverMetrics` -> `Metrics`, `PoolStatus` string -> enum.
Added a few new counter `acquiring`, `acquired`, `timedOutToAcquire`.
Changed to return a readonly dictionary so that a user cannot add/remove items in the returned dictionary.
Changed the ConnectionTimeHistogram to only record success connection time.
Changed the Acquisition Histogram to only record success acquisition time.


ali-ince (approved):
LGTM 👍

PR #278, opened 2018-03-07
Marshal enumerable to list

zhenlineo:
Supports both `Run(string, object)` and `Run(string, IDictionary<string, object>)`
When converting from object to `IDictionary<string, object>`, if the object implements `IEnumerable`, then it would call `GetEnumerator` rather than try to convert the class as a `IDictionary<string, object>`.


zhenlineo:
Closing this one as it will be implemented as Value types instead.

PR #279, opened 2018-03-15
Point support

ali-ince:
This PR brings `Point` data type support to .NET Driver.

As part of this work;
1. Made better use of `IPackStreamStructHandler` interface across bolt messages and internal value types,
2. Reworked and rebuilt `PackStream` and `Bolt` related unit tests, in order to make them less cryptic and easy to maintain,
3. Added V2 protocol support,
4. Updated `xUnit` version.

API that will expose creating `Point` instances is not finalised yet, so it is not in this PR.


Neo4j.Driver/Neo4j.Driver/Internal/IO/StructHandlers/PointHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+            return new Point(srId, x, y, z);
+        }
+
+        public void Write(IPackStreamWriter writer, object value)
+        {
+            switch (value)
+            {
+                case Point point:
+                {
+                    if (double.IsNaN(point.Z))
```
zhenlineo: Shall this testing the size of the value with `point.Dimension ` rather than NaN?

Neo4j.Driver/Neo4j.Driver/Internal/IO/StructHandlers/AckFailureMessageHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Neo4j.Driver.Internal.Messaging;
+using Neo4j.Driver.V1;
+
+namespace Neo4j.Driver.Internal.IO.StructHandlers
+{
+    internal class AckFailureMessageHandler : IPackStreamStructHandler
```
zhenlineo: Will it make sense if we have
```
internal abstract class WritableStructHandler : IPackStreamStructHandler
{
  public object Read(IPackStreamReader reader, byte signature, long size)
    {
      throw new ProtocolException($"It is not expected to receive {WritableTypes.toContentString()} messages from the server."); // or any meaning full way to address the type
    }
}
```
Similarly we could have a `ReadableStructHandler`.

BTW, We have a `toContentString` extension method to pretty print for `{string.Join(",", ReadableStructs)}`

Neo4j.Driver/Neo4j.Driver/V1/Types/IValue.cs
```diff
@@ -0,0 +1,24 @@
...
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Neo4j.Driver.V1
+{
+    public interface IValue
```
zhenlineo: Seems only Point is a `IValue` now. I assume `IValue` type is only for Cypher supported special Parameter types.
Pls add some docs for this new public API

zhenlineo: I do not know how to write this comment to class `ValueExtensions.cs`, so I am putting it here.

About `ValueExtensions.cs`, can you move the path of `ValueExtensions.cs`? it is in the public API. Will this enforce everyone to change the import path of this class?

zhenlineo (changes requested):
The tests looks good.
Just a few comments.

zhenlineo (approved):

PR #281, opened 2018-03-22
Introduce temporal types

ali-ince:
This PR adds support for temporal types. 

The following struct types are introduced as public API.

1. CypherDuration
2. CypherDate
3. CypherTime
4. CypherTimeWithOffset
5. CypherDateTime
6. CypherDateTimeWithOffset
7. CypherDateTimeWithZoneId

All types not ending with `Offset` or `ZoneId` represents local date/time values and the naming is chosen to keep type names consistent with their BCL counter-parts. For all types, if there's an available BCL type to be converted to, converter methods are available. If a truncation would occur during this conversion, a `TruncationException` would be thrown to protect data from accidental truncation.

The following functionality is not covered yet;
1. Sending BCL types directly to the server,
2. IConvertible implementations to support `Convert.ToDateTime` or `Convert.ChangeType` on these custom types,
3. IComparable implementations.

Based on #280.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/IO/StructHandlers/Temporal/DurationStructHandlerTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/IO/StructHandlers/Temporal/DurationStructHandlerTests.cs
@@ -29,9 +29,9 @@ using Xunit;
 
 namespace Neo4j.Driver.Tests.IO.StructHandlers
 {
-    public class CypherDurationStructHandlerTests : StructHandlerTests
+    public class DurationStructHandlerTests : StructHandlerTests
     {
-        internal override IPackStreamStructHandler HandlerUnderTest => new CypherDurationHandler();
+        internal override IPackStreamStructHandler HandlerUnderTest => new DurationHandler();
 
         [Fact]
         public void ShouldWriteDuration()
@@ -59,7 +59,7 @@ namespace Neo4j.Driver.Tests.IO.StructHandlers
             var writerMachine = CreateWriterMachine();
             var writer = writerMachine.Writer();
 
-            writer.WriteStructHeader(CypherDurationHandler.StructSize, CypherDurationHandler.StructType);
+            writer.WriteStructHeader(DurationHandler.StructSize, DurationHandler.StructType);
             writer.Write(21L);
             writer.Write(8L);
             writer.Write(564L);
--- a/Neo4j.Driver/Neo4j.Driver/Internal/IO/StructHandlers/Temporal/DurationHandler.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/IO/StructHandlers/Temporal/DurationHandler.cs
@@ -21,34 +21,36 @@ using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal.IO.StructHandlers
 {
-    internal class CypherDateTimeWithZoneIdHandler : IPackStreamStructHandler
+    internal class DurationHandler: IPackStreamStructHandler
     {
-        public const byte StructType = (byte) 'f';
-        public const int StructSize = 3;
+        public const byte StructType = (byte) 'E';
+        public const int StructSize = 4;
 
         public IEnumerable<byte> ReadableStructs => new[] {StructType};
 
-        public IEnumerable<Type> WritableTypes => new[] {typeof(CypherDateTimeWithZoneId)};
+        public IEnumerable<Type> WritableTypes => new[] {typeof(CypherDuration)};
 
         public object Read(IPackStreamReader reader, byte signature, long size)
         {
-            PackStream.EnsureStructSize("DateTimeWithZoneId", StructSize, size);
+            PackStream.EnsureStructSize("Duration", StructSize, size);
 
-            var epochSeconds = reader.ReadLong();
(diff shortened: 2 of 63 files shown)


Neo4j.Driver/Neo4j.Driver.Tests/IO/StructHandlers/Temporal/DurationStructHandlerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using Neo4j.Driver.Internal.IO;
+using Neo4j.Driver.Internal.IO.StructHandlers;
+using Neo4j.Driver.Internal.Messaging;
+using Neo4j.Driver.Internal.Types;
+using Neo4j.Driver.V1;
+using Xunit;
+
+namespace Neo4j.Driver.Tests.IO.StructHandlers
+{
+    public class DurationStructHandlerTests : StructHandlerTests
```
lutovich: Maybe just `DurationHandlerTests`? Similar tests for other temporal types do not have `Struct` in their name.

Neo4j.Driver/Neo4j.Driver/Internal/IO/StructHandlers/Temporal/DurationHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+        public IEnumerable<Type> WritableTypes => new[] {typeof(CypherDuration)};
+
+        public object Read(IPackStreamReader reader, byte signature, long size)
+        {
+            PackStream.EnsureStructSize("Duration", StructSize, size);
+
+            var months = reader.ReadLong();
+            var days = reader.ReadLong();
+            var seconds = reader.ReadLong();
+            var nanos = reader.ReadInteger();
```
lutovich, line 40: ~~Nanoseconds are actually `long` in the database (see `DurationValue` class). `#ReadLong()` should probably be used here~~

**Update:** nanoseconds are normilized in the database and can't have value greated than 999_999_999.

Neo4j.Driver/Neo4j.Driver/V1/Types/CypherDate.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        internal CypherDate(long epochDays)
+        {
+            EpochDays = epochDays;
+        }
+
+        /// <summary>
+        /// Days since Unix Epoch
+        /// </summary>
+        public long EpochDays { get; }
```
lutovich, line 59: I think it would be nice to expose year, month and day instead. This would make `CypherDate` useful even without additional libraries. Epoch days is kind of an implementation detail.

Neo4j.Driver/Neo4j.Driver/V1/Types/CypherDate.cs
```diff
@@ -0,0 +1,112 @@
...
+            return EpochDays.GetHashCode();
+        }
+
+        /// <summary>
+        /// Converts the value of the current <see cref="CypherDate"/> object to its equivalent string representation.
+        /// </summary>
+        /// <returns>String representation of this Point.</returns>
+        public override string ToString()
+        {
+            return $"Date{{epochDays: {EpochDays}}}";
```
lutovich, line 109: Could we instead return an ISO string?

Neo4j.Driver/Neo4j.Driver/V1/Types/CypherDateTimeWithOffset.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        /// <summary>
+        /// Fraction of seconds in nanosecond precision
+        /// </summary>
+        public int NanosOfSecond { get; }
+
+        /// <summary>
+        /// Offset in seconds precision
+        /// </summary>
+        public int OffsetSeconds { get; }
```
lutovich, line 128: Could `TimeSpan` be returned here?

Neo4j.Driver/Neo4j.Driver/V1/Types/CypherDuration.cs
```diff
@@ -0,0 +1,148 @@
...
+            Months = months;
+            Days = days;
+            Seconds = seconds;
+            Nanos = nanos;
+        }
+
+        /// <summary>
+        /// The number of months in this duration.
+        /// </summary>
+        public long Months { get; }
```
lutovich, line 82: We should probably consider exposing years and minutes in all drivers

lutovich (approved):
@ali-ince changes generally look good.

I think all new temporal types should expose meaningful fields instead of what we receive from the database. For example `CypherDate` can expose year, month and day instead of epoch day. It would also be great for `#ToString()` methods to return ISO formatted strings. Maybe Noda Time could be used for both and be only an internal dependency. This can be addressed in a separate PR.

~~Most important comment is about nanos being `long` in `CypherDuration`. Duration stored in the database is not necessarily normalized and can be like `5 months, 50 days, 100 seconds, 1000000000 nanos`.~~ **Update:** nanoseconds are normilized in the database and can't have value greated than 999_999_999.

ali-ince:
Thanks for the comments @lutovich.

I've addressed one of your comments for now (about class naming). I'm merging this for now and will work more on public API that we'll expose on `temporal types` and will create a new PR when it is finished.

PR #283, opened 2018-04-06
Marshal enumerable to list

zhenlineo:
Supports both `Run(string, object)` and `Run(string, IDictionary<string, object>)`
When converting from object to `IDictionary<string, object>`, if the object implements `IEnumerable`, then it would call `GetEnumerator` rather than try to convert the class as a `IDictionary<string, object>`.

Brings back https://github.com/neo4j/neo4j-dotnet-driver/pull/278


ali-ince:
LGTM.

PR #284, opened 2018-04-06
TimeZone issue when creating `CypherDateTime` from `DateTime`

zhenlineo:
By default `DateTime` uses `DateTimeKind.Unspecified` as the default value if not provided when creating.

However when calling `DateTime#ToLocalTime()` it will consider unspecified means `Utc` and convert from UTC to local.

However, when calling `DateTime#Equal`, the `Kind` will be ignored, and only Ticks will be compared.
This PR suggesting we should respect this `ToLocalTime` method think `Unspecified` as `Utc`. So added more docs for a user to read.


ali-ince:
👍

ali-ince (approved):
👍

PR #286, opened 2018-04-10
Expose individual temporal components in custom types

ali-ince:
This PR adds individual temporal components (year, month, day, hour, minute, second, nanosecond) to all corresponding temporal types and ensures that `ToString()` implementations generates `ISO 8601` compliant strings.

Date/Time to/from Epoch conversion code is adapted from https://github.com/ThreeTen/threetenbp.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Helpers/TemporalHelpers.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Helpers/TemporalHelpers.cs
@@ -201,12 +201,20 @@ namespace Neo4j.Driver.Internal
             }
         }
 
-        public static void AssertNoTruncation(long nanosecond, string target)
+        public static void AssertNoOverflow(IHasDateComponents date, string target)
         {
-            if (nanosecond % NanosecondsPerTick > 0)
+            if (date.Year > DateTime.MaxValue.Year || date.Year < DateTime.MinValue.Year)
+            {
+                throw new ValueOverflowException($"Year component ({date.Year}) of this instance is not valid for a {target} instance.");
+            }
+        }
+
+        public static void AssertNoTruncation(IHasTimeComponents time, string target)
+        {
+            if (time.Nanosecond % NanosecondsPerTick > 0)
             {
                 throw new ValueTruncationException(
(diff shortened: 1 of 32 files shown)


zhenlineo (commented):
Can we have a IT where we randomly generate some temporal types and bounce these values between server and client to ensure the value format is recognized the same between server and client?

JS driver has such a test where the client generate some dates that are not valid and get normalized at the server. So when the server returns the values back, they are different from what they were.

Neo4j.Driver/Neo4j.Driver/Internal/Helpers/TemporalHelpers.cs
```diff
@@ -16,92 +16,260 @@
...
+            var nanoOfDay = secondsOfDay * NanosPerSecond + nano;
+
+            ComponentsOfEpochDays(epochDay, out var year, out var month, out var day);
+            ComponentsOfNanoOfDay(nanoOfDay, out var hour, out var minute, out var second, out var nanosecond);
+
+            return new CypherDateTime(year, month, day, hour, minute, second, nanosecond);
         }
 
-        public static long NanosOf(int hour, int minute, int second, int nanoOfSecond)
+        private static long ComputeEpochDays(int year, int month, int day)
```
zhenlineo: @lutovich Can you also have a look at this method in this class from https://github.com/ThreeTen/threetenbp?

Neo4j.Driver/Neo4j.Driver/Internal/Helpers/TemporalHelpers.cs
```diff
@@ -16,92 +16,260 @@
...
+using Neo4j.Driver.Internal.Types;
 using Neo4j.Driver.V1;
 using TimeZoneConverter;
 
 namespace Neo4j.Driver.Internal
 {
     internal static class TemporalHelpers
     {
-        public const int NanosecondsPerTick = 100;
+        public const int MinYear = -999_999_999;
```
zhenlineo, line 30: @lutovich Can you also have a look at the new crazy methods added in this class?
Ref: https://github.com/ThreeTen/threetenbp

zhenlineo (commented):
Asking for external help...

PR #288, opened 2018-04-12
Implement IConvertible on temporal types

ali-ince:
This PR adds IConvertible implementations to each of the temporal types. Valid conversions are implemented using conversion properties on the actual types, and invalid conversions throw `InvalidCastException`.

Based on #287.


zhenlineo (commented):
Will interface be better compared to struct,
in the sense that we could only have the implementation once in a abstract class?

I would not make this commits blocking this PR. But if we could reduce some code duplications, then maybe we could expose the new types as interfaces after all in the renaming PR?

zhenlineo (approved):

PR #294, opened 2018-04-24
Embed CLDR time zone mapping table

ali-ince:
This PR embeds [CLDR time zone mapping table](http://cldr.unicode.org/development/development-process/design-proposals/extended-windows-olson-zid-mapping) that can be downloaded from [here](http://cldr.unicode.org/index/downloads/latest) into the driver to make IANA to windows time zone mapping possible.

Based on #296.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Temporal/TimeZoneMapping.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Temporal/TimeZoneMapping.cs
@@ -1,0 +1,161 @@
+﻿// Copyright (c) 2002-2018 "Neo Technology,"
+// Network Engine for Objects in Lund AB [http://neotechnology.com]
+// 
+// This file is part of Neo4j.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Temporal/TimeZoneMapping.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+            finally
+            {
+                SystemToTZInfo =
+                    new ReadOnlyDictionary<string, TimeZoneInfo>(TimeZoneInfo.GetSystemTimeZones()
+                        .ToDictionary(tz => tz.Id, tz => tz));
+            }
+        }
+
+        public static TimeZoneInfo Get(string zoneId)
```
zhenlineo, line 64: There are paths:
1) `if (SystemToTZInfo.TryGetValue(zoneId, out var tzInfo))`
2) `if (IANAToWindows.TryGetValue(zoneId, out var windowsId))`
3) `if (WindowsToIANA.TryGetValue($"{zoneId}_{GetCurrentTerritory()}", out var ianaId) ||`
4) `return TimeZoneInfo.FindSystemTimeZoneById(zoneId);`
5) error at step 4)

Could you make sure they are all covered by some unit tests to prevent anyone ever break this logic by mistake here?

The current unit tests looks only covers one path if I understand correctly?

ali-ince: @zhenlineo I'll add tests for other paths as another PR.

PR #295, opened 2018-04-25
Split metrics implementation

ali-ince:
This PR moves driver-level metrics implementation on another project named `Neo4j.Driver.Metrics` on its own, that will not be part of the nuget package and will leave the driver itself free of external dependencies.

Based on #294 and #296.


zhenlineo (approved):
The metrics part looks good. Nice idea to pull it out.

PR #297, opened 2018-04-25
Add additional tests for timezone mapping

ali-ince:
This PR adds time zone lookup tests that are applicable to unix environment.


ali-ince:
Replaced by #298.

PR #301, opened 2018-04-30
Couple of test improvements

ali-ince:
This PR tries to improve flakiness in some tests along with a main code fix that can cause a `NullReferenceException`.

* boltkit commands now timeout in 4 mins instead of 2 mins,
* metrics no more override existing metrics and atomically reads counters.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Metrics/ConnectionPoolMetrics.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Metrics/ConnectionPoolMetrics.cs
@@ -33,15 +33,15 @@ namespace Neo4j.Driver.Internal.Metrics
         private long _timedOutToAcquire;
 
         public int Creating => _creating;
-        public long Created => _created;
-        public long FailedToCreate => _failedToCreate;
+        public long Created => Interlocked.Read(ref _created);
+        public long FailedToCreate => Interlocked.Read(ref _failedToCreate);
 
         public int Closing => _closing;
-        public long Closed => _closed;
+        public long Closed => Interlocked.Read(ref _closed);
 
         public int Acquiring => _acquiring;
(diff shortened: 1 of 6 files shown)


Neo4j.Driver/Neo4j.Driver.Metrics/ConnectionPoolMetrics.cs
```diff
@@ -33,15 +33,15 @@ internal class ConnectionPoolMetrics : IConnectionPoolMetrics, IConnectionPoolLi
         private long _timedOutToAcquire;
 
         public int Creating => _creating;
-        public long Created => _created;
-        public long FailedToCreate => _failedToCreate;
+        public long Created => Interlocked.Read(ref _created);
```
zhenlineo, line 36: To me, I do not feel a need to read a very precise value for metrics. If the values in the test assertion are sometimes wrong, then maybe we shall just remove that assertion in the tests? 

I feel it might be a big performance impact if we always atomically read these values for metrics.

ali-ince: This was a spike to see if there's some kind of concurrency issue there. I will continue working to hunt for a connection leakage - but I agree that we can revert this commit in all.

zhenlineo: Then why not also have Creating and others also to be read atomically?

ali-ince: Reading of `int` (`Int32`) values are already atomic, so we do not need to `Interlocked.Read` them and that's the reason why we don't have an overload of `Interlocked.Read` for `int` values.

PR #303, opened 2018-05-08
Fix nanosecond normalisation

ali-ince:
This PR fixes the following aspects of nanosecond handling;

1. It no more prints nanoseconds part if it is `0` on each temporal type.
2. It normalises second part based on its sign and nanosecond value while creating the ISO 8601 compatible string representation.

The `Duration` class restricts nanosecond field to be between `0` and `999.999.999` and differs from the js driver in that respect. What do you think @lutovich & @zhenlineo?


Neo4j.Driver/Neo4j.Driver.Tests/Types/DurationTests.cs
```diff
@@ -74,8 +74,11 @@ public void ShouldCreateDurationWithDaysSecondsAndNanoseconds()
         [InlineData(0, 32, 785, 789215800, "P0M32DT785.789215800S")]
         [InlineData(0, 0, 785, 789215800, "P0M0DT785.789215800S")]
         [InlineData(0, 0, 0, 789215800, "P0M0DT0.789215800S")]
-        [InlineData(0, 0, 0, 0, "P0M0DT0.000000000S")]
-        [InlineData(500, 0, 0, 0, "P500M0DT0.000000000S")]
+        [InlineData(0, 0, -1, 0, "P0M0DT-1S")]
+        [InlineData(0, 0, 0, 999999999, "P0M0DT0.999999999S")]
+        [InlineData(0, 0, -1, 5, "P0M0DT0.999999995S")]
+        [InlineData(0, 0, -1, 999999999, "P0M0DT0.000000001S")]
```
zhenlineo: Shall this be `P0M0DT-0.000000001S`?

Neo4j.Driver/Neo4j.Driver.Tests/Types/DurationTests.cs
```diff
@@ -74,8 +74,11 @@ public void ShouldCreateDurationWithDaysSecondsAndNanoseconds()
         [InlineData(0, 32, 785, 789215800, "P0M32DT785.789215800S")]
         [InlineData(0, 0, 785, 789215800, "P0M0DT785.789215800S")]
         [InlineData(0, 0, 0, 789215800, "P0M0DT0.789215800S")]
-        [InlineData(0, 0, 0, 0, "P0M0DT0.000000000S")]
-        [InlineData(500, 0, 0, 0, "P500M0DT0.000000000S")]
+        [InlineData(0, 0, -1, 0, "P0M0DT-1S")]
+        [InlineData(0, 0, 0, 999999999, "P0M0DT0.999999999S")]
+        [InlineData(0, 0, -1, 5, "P0M0DT0.999999995S")]
```
zhenlineo: Shall this be `P0M0DT-0.999999995S` as the result should be `-1 + 0.000000005S`?

zhenlineo (commented):
So after the normalization on the server side,
the value of nanosecond shall always be in the range 0~999 999 999. 

Also when we call Duration.Nanos then we shall get a value in the range of 0~999 999 999.
But when we call Duration.ToNanos then we shall get a value includes seconds + nanos.

ali-ince:
@zhenlineo missing sign issue should now be addressed.

PR #305, opened 2018-05-14
Introduce trust managers

ali-ince:
This PR changes the trust mechanism how Driver establishes trust during TLS handshake. Now it is possible to completely override default `TrustManager` behaviour and provide a custom one based on possibly different security policies.

`Config.TrustStrategy` / `IConfigBuilder.WithTrustStrategy` is now obsolete and `Config.TrustManager` / `IConfigBuilder.WithTrustManager` is the mechanism to change the underlying trust manager that will be used.

There are 4 different trust managers that's provided with the driver (they can be created through `TrustManager` class);

1. `InsecureTrustManager`: It trusts any certificate and can be configured to perform a hostname verification. This is currently the default trust manager and will be changed in upcoming versions.
2. `ChainTrustManager`: It uses operating system's system/user root certificate stores to establish the trust. It can be configured to enable/disable hostname verification, revocation check and whether to use machine context instead of user's.
3. `PeerTrustManager`: It uses operating system's system/user TrustedPeople certificate stores to establish the trust. It can be configured to enable/disable hostname verification and whether to use machine context instead of user's.
4. `CertificateTrustManager`: It trusts certificates that are identical / issued by the provided certificates. It can be configured to enable/disable hostname verification.


Neo4j.Driver/Neo4j.Driver/V1/TrustManager.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="sslPolicyErrors">The initial policy errors that shows what problems were detected during the handshake</param>
+        /// <returns><value>true</value> if the connection should be established, <value>false</value> otherwise</returns>
+        public abstract bool ValidateServerCertificate(Uri uri, X509Certificate2 certificate, X509Chain chain,
+            SslPolicyErrors sslPolicyErrors);
+
+        /// <summary>
+        /// Creates an insecure trust manager that trusts any certificate it is presented, but does hostname verification.
+        /// </summary>
+        /// <returns>An instance of <see cref="TrustManager"/></returns>
+        public static TrustManager CreateInsecure() => CreateInsecure(false);
```
zhenlineo: Seems different from docs. The hostname verification is turned off by default for this one.
Not sure if we shall turn it on by default.
Maybe worth to test on linux and macOs machines too?

zhenlineo: Is this feature fully available on all platforms?

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
```diff
@@ -84,7 +84,7 @@ public void ShouldBlockWhenMaxPoolSizeReached()
 
                 pool.NumberOfIdleConnections.Should().Be(0);
                 pool.NumberOfInUseConnections.Should().Be(2);
-                timer.ElapsedMilliseconds.Should().BeGreaterOrEqualTo(1000);
+                Math.Round(timer.Elapsed.TotalSeconds).Should().BeGreaterOrEqualTo(1);
```
zhenlineo: Is it correct to have Math round over it? Shouldn't it just greater than 1?
I am thinking if it is enough to only have `timmer.Elapsed.TotalSeconds.Should().BeGreaterOrEqualTo(1);`?

zhenlineo (commented):
LG

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
ali-ince: The one you suggested was failing for some other tests - the problem lies on the fact that `timer.Elapsed.TotalSeconds` reports slightly less than `1` (i.e. `0.998...`) mostly on our brand new linux build agents. 

It is similar to the other elapsed time assertions we made on this class - I changed all of them to be rounded to the nearest integral. They all have the same pattern that sometimes they fail with slightly less elapsed times than expected, which I believe is a result of timer resolution on different platform, hardware and/or operating system.

zhenlineo: Would it be to inaccurate for this number? Maybe `Math.Round(timer.Elapsed.TotalSeconds*10).Should().BeGreaterOrEqualTo(10)`? Or put in a new method such as 
```
double round(double value, double accuracy) {
  return Math.Round(value/accuracy)*accuracy;
}
round(timer,Elapsed.TotalSeconds, 0.1).Should().BeGreaterOrEqualTo(1);
```

ali-ince: I got rid of `Math.Round` calls and instead add an overloaded `BeGreaterOrEqualTo` assertion that takes an accuracy parameter.

PR #308, opened 2018-06-15
Adding tests for failure to reset message

zhenlineo:


Neo4j.Driver/Neo4j.Driver.IntegrationTests/BoltStubServerTests.cs
```diff
@@ -83,5 +83,26 @@ public void CanSendMultipleBookmarks()
                 }
             }
         }
+
+        [RequireBoltStubServerFact]
+        public void ShouldCloseConnectionWhenFailingToReset()
```
ali-ince, line 88: I'm not sure if I got this test correct but since we are using two consecutive runs of `BoltStubServer` and we can't expect the to use the first connection on the second since the first server is already shutdown. Am I missing something here?

ali-ince (commented):
Wdyt about verifying this behaviour also with some unit tests?

Neo4j.Driver/Neo4j.Driver.IntegrationTests/BoltStubServerTests.cs
zhenlineo: I actually agree with you. I will re-think of this test.

PR #309, opened 2018-06-20
Modify the source code to compile and run tests on mac with VS2017

zhenlineo:
TCK tests are not supported and need to be excluded from the project


zhenlineo:
Included in #310

PR #314, opened 2018-08-07
Refactoring to get ready for new BoltV3

zhenlineo:
Enhance the `BoltProtocol` to be able to handle session.run and tx methods directly.

After this refactoring, to add BoltProtocolV3, we will just need to implements `IBoltProtocol` and `MessageFormat` for BoltV3.

Based on #313


ali-ince (approved):
:+1:

PR #317, opened 2018-08-27
Boltv3

zhenlineo:
Added new protocol BoltV3.
In this new protocol, we enabled run with metadata. Therefore we made it possible to 1) include `Session#Run` into causal chain (bookmark), 2) passing transaction timeout and transaction metadata (via TransactionConfig) at both `Session#Run` and `Session#BeginTransaction`.

New API:
`TransactionConfig`
`Session.Run` and `Session.BeginTransaction` and Session run transaction with built-in retry methods with extra param `TransactionConfig`

Usage Example:
```
var txConfig = new TransactionConfig {Metadata = new Dictionary<string, object> {{"name", "Molly"}}};
using (var session = Server.Driver.Session())
{
  session.Run("CREATE (:Node)", txConfig).Consume();
  session.Run("MATCH (n:Node) RETURN n", txConfig).Single();
  // causal chaining within a session ensures to see the node just created.
}
```


Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV1MessageFormat.cs
```diff
@@ -23,6 +23,22 @@ namespace Neo4j.Driver.Internal.Protocol
...
+        public const byte MsgRun = 0x10;
+        public const byte MsgDiscardAll = 0x2F;
+        public const byte MsgPullAll = 0x3F;
+
+        public const byte MsgRecord = 0x71;
+        public const byte MsgSuccess = 0x70;
+        public const byte MsgIgnored = 0x7E;
+        public const byte MsgFailure = 0x7F;
+
+        #endregion Consts
```
lutovich: Is "Consts" needed here?

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV3MessageFormat.cs
```diff
@@ -0,0 +1,78 @@
...
+    internal class BoltProtocolV3MessageFormat: MessageFormat
+    {
+        #region Message Constants
+
+        public const byte MsgHello = 0x01;
+        public const byte MsgBegin = 0x11;
+        public const byte MsgCommit = 0x12;
+        public const byte MsgRollback = 0x13;
+
+        #endregion Consts
```
lutovich: Also maybe remove "Consts"?

lutovich (approved):

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV1.cs
```diff
@@ -172,5 +176,21 @@ public void Reset(IConnection connection)
         {
             connection.Enqueue(ResetMessage.Reset, null);
         }
+        
+        private void AssertNullOrEmptyTransactionConfig(TransactionConfig txConfig)
```
ali-ince: I'm not sure about this but would it ever be possible to run 3.4 and 3.5 mixed - hit this assertion in this mixed version environment? What about simply ignoring `txConfig` (and warning about it) when connected to `V2` and `V1` - also address this in docs?

Neo4j.Driver/Neo4j.Driver/Internal/Session.cs
```diff
@@ -53,36 +54,65 @@ internal partial class Session : StatementRunner, ISession
...
+        public override IStatementResult Run(Statement statement)
+        {
+            return Run(statement, TransactionConfig.Empty);
+        }
+        public override Task<IStatementResultCursor> RunAsync(Statement statement)
+        {
+            return RunAsync(statement, TransactionConfig.Empty);
+        }
+
+        public ITransaction BeginTransaction(TransactionConfig txConfig=null)
```
ali-ince: Shall we extract method overloads instead of default parameters (for better C++/CLI integration) in the rest of the class (for `public` methods)?

Neo4j.Driver/Neo4j.Driver/V1/ISession.cs
```diff
@@ -47,16 +47,10 @@ public interface ISession : IStatementRunner
...
-        ITransaction BeginTransaction();
-
-        /// <summary>
-        /// This method is deprecated.
-        /// </summary>
-        /// <param name="bookmark">A bookmark.</param>
-        /// <returns>A new transaction.</returns>
-        [Obsolete("Pass in bookmark at session instead.")]
-        ITransaction BeginTransaction(string bookmark);
+        ITransaction BeginTransaction(TransactionConfig txConfig=null);
```
ali-ince: Shall we extract method overloads instead of default parameters (for better C++/CLI integration) in the rest of the interface?

ali-ince (commented):
@zhenlineo I made a couple of comments. I think removing optional parameters would make sense to keep   our C++/CLI support at its current state (we previously removed optional parameters).

PR #320, opened 2018-09-05
Goodbye message

zhenlineo:
Based on #317 

When a new connection established, we "login" by sending an Init/Hello message.
When closing a connection, we "logout" by sending a Goodbye message. 
Any error happened when sending Goodbye message will not be surfaced but only logged on debug level.


zhenlineo:
See https://github.com/neo4j/neo4j-dotnet-driver/pull/321

PR #321, opened 2018-09-06
Added goodbye message for BoltV3

zhenlineo:
Based on #317 

When a new connection established, we "login" by sending an Init/Hello message.
When closing a connection, we "logout" by sending a Goodbye message. 
Any error happened when sending Goodbye message will not be surfaced but only logged on debug level.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
@@ -70,7 +70,7 @@ namespace Neo4j.Driver.Internal.Connector
             try
             {
                 _boltProtocol = _client.Connect();
-                _boltProtocol.Authenticate(this, _userAgent, _authToken);
+                _boltProtocol.Login(this, _userAgent, _authToken);
             }
             catch (AggregateException e)
             {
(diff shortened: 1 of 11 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -171,7 +171,13 @@ public void Reset()
 
         public bool IsOpen => _client.IsOpen;
         public IServerInfo Server { get; set; }
-        public IBoltProtocol BoltProtocol => _boltProtocol;
+
+        public IBoltProtocol BoltProtocol
```
ali-ince, line 175: Do we need this change, as far as I see, we're only reading this from outside?

ali-ince (commented):
LGTM. I only had one small comment about a property.

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
zhenlineo: It was used for test only. I want to pass in a mocked protocol. Without a setter, the only way to set it was via a success Conn.Init. As this class is internal so I directly added a setter.

PR #322, opened 2018-09-07
Custom resolver function to provide routing driver initial server addresses

zhenlineo:


Neo4j.Driver/Neo4j.Driver/Internal/RoutingSettings.cs
```diff
@@ -17,24 +17,55 @@
...
+
+            InitialServerAddressProvider = new InitialServerAddressProvider(initServerUri, config.Resolver);
             RoutingContext = routingContext;
             Strategy = config.LoadBalancingStrategy;
         }
     }
+
+    internal interface IInitialServerAddressProvider
+    {
+        ISet<Uri> Resolve();
```
lutovich: It's a bit strange to have a "provider" with "resolve" methods. Wdyt about renaming methods to `Get` and `GetAsync`?

lutovich (commented):
Looks good. Made a small comment. It would also be nice to add a stub server test which tests the custom resolver.

Neo4j.Driver/Neo4j.Driver/V1/IAddressResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <param name="uri">The server address to resolve.</param>
+        /// <returns>Resolved server addresses in a set.</returns>
+        ISet<Uri> Resolve(Uri uri);
+
+        /// <summary>
+        /// Given a server address, asynchronously resolves the address and returns the resolved server addresses in a set.
+        /// </summary>
+        /// <param name="uri">The server address to resolve.</param>
+        /// <returns>A task of resolved server addresses in a set.</returns>
+        Task<ISet<Uri>> ResolveAsync(Uri uri);
```
ali-ince: Do we really need an async version? I believe mostly people will use it to return a static list of hosts, won't they?

Neo4j.Driver/Neo4j.Driver/V1/IAddressResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public interface IServerAddressResolver
+    {
+        /// <summary>
+        /// Given a server address with host name and port defined in <see cref="Uri"/>,
+        /// returns the resolved server addresses with host name and port saved in a set of <see cref="Uri"/>.
+        /// </summary>
+        /// <param name="uri">The server address to resolve.</param>
+        /// <returns>Resolved server addresses in a set.</returns>
+        ISet<Uri> Resolve(Uri uri);
```
ali-ince: How do you feel about not using `Uri` for this? We only want users to map `hostname:port` pairs to another set of `hostname:port` pairs and `Uri` contains more information than that. Wdyt?

ali-ince (commented):
Looks good. Made a couple of comments about the new public `IServerAddressResolver` interface.

Neo4j.Driver/Neo4j.Driver/V1/IAddressResolver.cs
zhenlineo: Do you think this async method would be useful if a user uses some DNS resolver or server address registry service to tell what servers are available? If you could only think of static use cases, then we definitely do not need a async version.

ali-ince: Yeah, maybe - but I believe it will be static most of the time. We can remove the async one here and consider adding it later if anyone asks for it?

Neo4j.Driver/Neo4j.Driver/V1/IAddressResolver.cs
```diff
@@ -29,32 +29,113 @@ namespace Neo4j.Driver.V1
...
         /// <returns>Resolved server addresses in a set.</returns>
-        ISet<Uri> Resolve(Uri uri);
+        ISet<ServerAddress> Resolve(ServerAddress address);
+    }
+
+    /// <summary>
+    /// A server address consists of <see cref="ServerAddress.Host"/> and <see cref="ServerAddress.Port"/>.
+    /// This address specifies where the driver to find the server.
+    /// </summary>
+    public sealed class ServerAddress
```
ali-ince: It may worth adding `IEquatable<ServerAddress>` as an implemented interface.

PR #327, opened 2018-10-08
Add custom resolver example

ali-ince:


zhenlineo (commented):
It would be really really appreciate if you could write this test with a stub server. As it is really slow to start a cluster.

ali-ince:
@zhenlineo I've changed the test to use stub server.

PR #328, opened 2018-11-02
Allow Transaction to be used remotely

simendsjo:
I won't pretend to know exactly why this works or if there are better
ways to solve the issues, so take what I state here with a grain of
salt.

The Neo4jClient driver supports TransactionScope and promotion of a
transaction to a distributed transaction. TransactionScope is important
when using other transactions as well, e.g. SQL transactions.

When starting a SQL transaction ofter a Neo4j transaction is started,
the transaction have to be promoted to a distributed transaction.
Neo4jClient uses AppDomain to handle this (at least I think that's the
reason), and AppDomain requires MarshalByRefObject to work properly.

Without MarshalByRefObject, things just doesn't work. When trying to
debug why nothing gets commited, the debugger breaks with a message that
the Neo4j Transaction isn't serializable. Adding SerializableAttribute
wasn't possible as this was needed all the way down, and
ConcurrentDictionary (or something similar) wasn't marked as
Serializable.

Adding MarshalByRefObject fixes Neo4jClient, allowing the Bolt
transactions to be commited.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/StatementRunner.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/StatementRunner.cs
@@ -23,7 +23,15 @@ using Neo4j.Driver.V1;
 
 namespace Neo4j.Driver.Internal
 {
-    internal abstract class StatementRunner : IStatementRunner
+    internal abstract class StatementRunner :
+#if NET452
+        // Needed to use a Transaction remotely.
+        // This can be used by drivers to add a Transaction to a custom AppDomain
+        // to implement promotion to a distributed transaction (as done by
+        // the Neo4jClient driver).
+        MarshalByRefObject,
+#endif
+        IStatementRunner
     {
         public abstract IStatementResult Run(Statement statement);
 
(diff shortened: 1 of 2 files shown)


simendsjo:
Anyone here? I notice the other PR in the project also fails the tests, but the TC instance isn't public, so I cannot see what fails.

zhenlineo:
@simendsjo Sorry I would like to talk this PR with @ali-ince. Then I forgot about this.

ali-ince:
I've looked at this and also the latest changes committed into neo4jclient, and I'm not sure we really need to make `Transaction` a `MarshalByRefObject`. I would like to verify a couple of things with @cskardon before going forward.

Sorry for keeping you waiting @simendsjo, I'll update whenever we have some progress.

P.S. Don't worry about TC failures for the time being, they'll not be built until you sign our [CLA](https://neo4j.com/developer/cla/).

simendsjo:
> I've looked at this and also the latest changes committed into neo4jclient, and I'm not sure we really need to make Transaction a MarshalByRefObject. I would like to verify a couple of things with @cskardon before going forward.

Any update on this?

ali-ince:
Hi @simendsjo,

Spending more time on this, we're against adding `MarshalByRefObject` to the driver source code. This is basically because;

* it increases code complexity for a feature we're not implementing and testing (and we currently don't have a plan to support `System.Transaction` on the driver),
* it's the upstream project's responsibility to manage the marshal by reference logic if the underlying objects are not serializable (which is not possible at all for transient objects like session, transaction, etc.).

I'm closing this and suggest you to take this to Neo4jClient with above justifications.

Thanks.

PR #331, opened 2018-12-21
Removed TCK and replaced with normal ITs

zhenlineo:
Added a system variable `DOTNET_DRIVER_USING_LOCAL_SERVER=true` to directly run ITs with a local single instance

(This PR shall be re-targeted at 2.0 branch)


ali-ince (approved):
LGTM :+1:

PR #336, opened 2019-02-11
Build synchronous API on asynchronous calls

ali-ince:
.NET Driver included two distinct code paths for synchronous and asynchronous API. This was an explicit decision at the time of introducing async API with 1.5. Given that this dual code path increases the code complexity, `Main` methods can be asynchronous (with C# 7.3) - and most of the frameworks already supporting asynchronous methods - it may be a good time to remove the synchronous code path across the driver implementation and build it on top of the asynchronous API.

This PR introduces a first phase of changes towards this direction.


Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscoveryManager.cs
```diff
@@ -52,35 +57,10 @@ public ClusterDiscoveryManager(IConnection connection, IDictionary<string, strin
...
-                    $"Invalid discovery result: discovered {Routers.Count()} routers, " +
-                    $"{Writers.Count()} writers and {Readers.Count()} readers.");
-            }
-        }
-
         public async Task RediscoveryAsync()
         {
             var provider = new SingleConnectionBasedConnectionProvider(_conn);
-            var session = new Session(provider, _logger);
+            var session = new Session(provider, _logger, _syncExecutor);
```
zhenlineo, line 63: Seems this is not necessary needed as we are only using the async part. Maybe a dummy executor which will throw exception if we try to do sync operation here is enough. I just feel a lot of effort to pass this `_syncExecutor` down but we actually do not need it.

zhenlineo (approved):
Added some ideas about possible improvement.

PR #339, opened 2019-05-01
Make retries to always give a second chance

ali-ince:
This PR changes the retry logic so that a failed (and retryable) action will be retried at least another time even if the first run of the action took more time than the configured maximum retry duration.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/RetryLogic.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/RetryLogic.cs
@@ -58,9 +58,12 @@ namespace Neo4j.Driver.Internal
         {
             var exceptions = new List<Exception>();
             var timer = new Stopwatch();
-            timer.Start();
+            var delay = TimeSpan.Zero;
             var delayMs = _initialRetryDelayMs;
             var retryCount = 0;
+            var shouldRetry = false;
+
+            timer.Start();
             do
             {
                 retryCount++;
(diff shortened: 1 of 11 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/RetryLogic.cs
```diff
@@ -70,14 +69,25 @@ public T Retry<T>(Func<T> runTxFunc)
...
+                    }
 
-                    var delay = TimeSpan.FromMilliseconds(ComputeDelayWithJitter(delayMs));
-                    _logger?.Warn(e, $"Transaction failed and will be retried in {delay}ms.");
-                    Thread.Sleep(delay);
-                    delayMs = delayMs * _multiplier;
+                    exceptions.Add(e);
+                    break;
                 }
-            } while (timer.Elapsed.TotalMilliseconds < _maxRetryTimeMs);
```
zhenlineo, line 80: I found it is easier to read with this do and while loop than the loop with `continue` and `break`. It looks to me the reason the new code did another run because the check of `timer.ElapsedMilliseconds` happens before sleep. So is it possible that we just keep a var before sleep and compare the var here?

zhenlineo (approved):
The logic looks good, just a small code style suggestion.

Neo4j.Driver/Neo4j.Driver/Internal/RetryLogic.cs
ali-ince: Changed the code to mostly its previous version and added a simple check to the `do-while` loop. This looks simpler, thanks!

PR #346, opened 2019-07-02
Split session API

ali-ince:
The Session API is now split into three different interfaces;
1. `IDriver.AsyncSession()`: can be used to acquire asynchronous sessions, which will be the only implementation provided by the `Neo4j.Driver` assembly / nuget package.
2. `IDriver.Session()`: can be used to acquire synchronous sessions, which are built on top of the asynchronous sessions and could only be used when `Neo4j.Driver.Simple` assembly / nuget package is referenced.
3. `IDriver.RxSession()`: can be used to acquire reactive sessions, which are built on top of the asynchronous sessions and could only be used when `Neo4.Driver.Reactive` assembly / nuget package is referenced.

Based on #345.


Neo4j.Driver/Neo4j.Driver.Reactive/Internal/InternalRxSession.cs
```diff
@@ -67,7 +67,7 @@ public IRxResult Run(string statement, object parameters, TransactionConfig txCo
         public IRxResult Run(Statement statement, TransactionConfig txConfig)
```
zhenlineo: Suggest to rename `IRxResult` to `IRxStatementResult` to be aligned with `IStatementResult` and `IStatementResultCursor`

zhenlineo (approved):
It looks pretty good. All the classes are in a good place and connected via `AsyncSession` and `StatementResultCursor` in a nice way. 

`BlockingExecutor` is a bit magic to me. It is not straightforward to me if there is any deadlock problem to scheduled blocking operations when chained together. It looks like we will chain all network reading, writing, our own driver logic together. I assuming it was the same as the blocking API before.

ali-ince:
@zhenlineo `BlockingExecutor` is actually not a magic and it only executes the async operations in a task that has assigned default options. It's there in case we need to tweak it more over time.

There's still the possibility to run into thread starvation and in environments where it is expected to have several concurrent requests, synchronous API is definitely not preferred and asynchronous or reactive API should be used.

PR #362, opened 2019-11-08
Updated driver metrics

zhenlineo:
Removed connection metrics and removed the use of Histogram


zhenlineo:
Consider making this into public API

PR #384, opened 2020-01-16
Clean-up the project files and add support for Sourcelink.

teo-tsirpanis:
This PR adds support for Sourcelink (the symbol packages must also be uploaded to NuGet), cleans-up the project files by removing redundant package references that are already included in .NET Standard 2.0. More clean-ups are better explained in the commit messages.


zhenlineo:
`CertificateTrustIT` tests are broken on my test agents. We need to look into this PR a bit more. As it also requires some tooling updates in our test agents.

teo-tsirpanis:
@zhenlineo, why not make the TeamCity server publicly accessible? It would greatly help situations like this one. Unless there are confidential information of course.

PR #391, opened 2020-03-05
Create a database selection example.

michael-simons:
See neo4j/neo4j-java-driver#684 and https://github.com/neo4j/neo4j-javascript-driver/pull/533

Was not yet able to run this locally due to issues with Boltkit and Environment variables, so this might need some polishing.


michael-simons:
It seems to me the same tests as with JavaScript fail.

zhenlineo:
Merged!

PR #447, opened 2020-08-28
Do not close testkit connection upon driver error

2hdddg:
It is legal for testkit frontend to generate errors in driver,
connection should not be closed after that.

Only two types of errors supported by frontend, BackendError and
DriverError. Sending anything else from backend will cause exceptions on
the frontend, remap all neo4j errors to DriverError, the rest to
BackendError.


AndyHeap-NeoTech (commented):
Looks good.

PR #466, opened 2021-01-12
Don't force resuming continuation on same thread

simendsjo:
I assume these can be continued on other threads. If not, a comment should be added instead.


AndyHeap-NeoTech:
You are correct, well spotted. Thanks.

PR #492, opened 2021-02-19
Improved output messages to be clearer.

AndyHeap-NeoTech:
This will highlight if there is an empty message.


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Reactive/NestedQueriesIT.cs
```diff
@@ -78,8 +78,8 @@ public void ShouldCloseSessionWhenRunNestedQueriesWithSessionRuns()
 
 		public bool OutputMessage(string message, string expectedMessage)
 		{
-			Output.WriteLine(message);
-			return message.Contains(expectedMessage);
+			Output.WriteLine("Actual Message: " + message);
+			return message.Contains("Excpeted Message: " + expectedMessage);
```
fbiville:
```suggestion
			return message.Contains("Expected Message: " + expectedMessage);

```

PR #525, opened 2021-05-10
testkit-backend: Writing error over the wire when ArgumentException occurs

bigmontz:
This issue was not allowing to run the tests to check if the driver will fail to be created when there are url query-string params for a bolt (direct) connection.


AndyHeap-NeoTech (approved):
Good find, surprised it has not become apparent before now.

PR #553, opened 2021-09-08
Add basic logger to drivers created in TestKit backend

robsdedude:


AndyHeap-NeoTech (approved):
An item I can now remove from my To-Do list, thanks :) Do we need a more comprehensive logger implementation going forward?

robsdedude:
Optionally, the logger could send the messages to the TestKit. This makes it easier to debug failing tests if you're running the .NET driver in docker for instance.
But that's more of a nice to have than a priority. I wanted to talk about this option in an upcoming knowledge sharing session.

AndyHeap-NeoTech:
> Optionally, the logger could send the messages to the TestKit. This makes it easier to debug failing tests if you're running the .NET driver in docker for instance.
> But that's more of a nice to have than a priority. I wanted to talk about this option in an upcoming knowledge sharing session.

I like this idea.

PR #554, opened 2021-09-08
Fix parameter decoding in backend's tx run

robsdedude:
This backend bug currently blocks https://github.com/neo4j-drivers/testkit/pull/240


AndyHeap-NeoTech (approved):
Well spotted. Looks like it was just missed when the conversions were added. What lead you to notice it?

robsdedude:
I changed some integration tests in TestKit to use a transaction function to be cluster-safe. Then they began to fail for .NET with the weirdest driver behavior :upside_down_face:

PR #555, opened 2021-09-09
Implement consume on transaction results

robsdedude:
TransactionResults in TestKit backend had no consume method.

Found as it failed in (and currently blocks) https://github.com/neo4j-drivers/testkit/pull/240

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Result/Result.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Result/Result.cs
@@ -10,6 +10,9 @@ namespace Neo4j.Driver.Tests.TestBackend
 {
 	internal class Result : IProtocolObject
 	{
+		[JsonIgnore]
+		public IResultCursor ResultCursor { get; set; }
+
 		public ResultType data { get; set; } = new ResultType();
 
         public class ResultType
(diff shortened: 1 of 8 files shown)


Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Result/Result.cs
```diff
@@ -55,6 +60,11 @@ public async Task PopulateRecords(IResultCursor cursor)
 			ResultCursor = cursor;
 			await Task.CompletedTask;
 		}
+
+		public async override Task<IResultSummary> ConsumeResults()
```
AndyHeap-NeoTech: As the two implementations are the same I think it would be cleaner to put the ConsumeResults into the base Result class as a public method and remove it from the two derived classes. Inheritance would take care of the rest.

Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Result/Result.cs
```diff
@@ -73,7 +83,7 @@ public async override Task<IRecord> GetNextRecord()
 			return await Task.FromResult<IRecord>(null);
 		}
 
-		public async Task<IResultSummary> ConsumeResults()
+		public async override Task<IResultSummary> ConsumeResults()
```
AndyHeap-NeoTech: See previous comment about inheritance

AndyHeap-NeoTech (approved):

PR #567, opened 2022-01-12
Proper handling of both zero and null tx timeout

thelonelyvulpes:
Change tx timeout to allow users to pass 
- `null` (default server timeout).
- `TimeSpan.Zero` (No timeout).


Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Session/SessionRun.cs
```diff
@@ -25,32 +24,29 @@ public class SessionRunType
...
 
         void TransactionConfig(TransactionConfigBuilder configBuilder)
         {
-            if (data.timeout != -1)
+            if (data.timeout.HasValue)
             {
-                var time = TimeSpan.FromMilliseconds(data.timeout);
-                configBuilder.WithTimeout(time);
+                configBuilder.WithTimeout(TimeSpan.FromMilliseconds(data.timeout.Value));
             }
```
robsdedude: I don't think this is doing what I envisioned it to do. There should be three cases covered here:
 * Set the timeout to `null` explicitly
 * Set the timeout to a given time span
 * Don't set the timeout at all

This code never runs into the first case.

robsdedude (changes requested):

robsdedude (approved):

PR #584, opened 2022-02-04
Test backend encryption check

thelonelyvulpes:
implement test backend for checking if encrypted with naïve encryption check


AndyHeap-NeoTech (approved):
Looks good. That storing of the Uri for just testing is irritating though. We should revisit that in the future :)

PR #586, opened 2022-02-10
Updated comment to match the default value for MaxConnectionPoolSize

AndyHeap-NeoTech:
After changing the MaxConnectionPoolSize value to a default of 100 from 500 the associated comment was missed.


thelonelyvulpes (approved):
:+1:

PR #588, opened 2022-02-17
Socket acquisition timeouts

thelonelyvulpes:


AndyHeap-NeoTech (approved):
The solution looks good. I particularly like the introduction of the Task extension methods to take a timeout and also handle a CancellationToken correctly. An elegant solution that nicely expands on the original stream extension I added.

PR #589, opened 2022-02-25
fix for possible resource leak

thelonelyvulpes:
* fix when we have a -1 connection read timeout the Delay timeout task can exist for the scope of the application.


AndyHeap-NeoTech (approved):
I really like the solution.

PR #590, opened 2022-02-28
Refactor timeout exception into timeout

thelonelyvulpes:


AndyHeap-NeoTech (approved):
Very happy with this, as discussed in our meeting.

PR #592, opened 2022-03-04
fix error message being overridden by exception

thelonelyvulpes:
fix Connection acquisition timeout being reported with the exception instead of how we want it surfaced.


Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/TestBlackList.cs
```diff
@@ -89,7 +89,8 @@ static class TestBlackList
 			("test_routing_v3.RoutingV3.test_should_successfully_send_multiple_bookmarks",
 				"Test failing requires investigation"),
 
-
```
AndyHeap-NeoTech, line 92: Has this one been flaky for long?

AndyHeap-NeoTech (approved):
Nice and easy, my favourite type of review :)

PR #594, opened 2022-03-07
Update discovery exceptions & handling

thelonelyvulpes:
- surface error codes from discovery
- enable testkit feature flag


Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV4_0.cs
```diff
@@ -154,7 +154,34 @@ await connection
         protected internal override void GetProcedureAndParameters(IConnection connection, string database, out string procedure, out Dictionary<string, object> parameters)
         {
             procedure = GetRoutingTableForDatabaseProcedure;
-            parameters = new Dictionary<string, object> { { "context", connection.RoutingContext }, { "database", string.IsNullOrEmpty(database) ? null : database } };                     
+            parameters = new Dictionary<string, object> { { "context", connection.RoutingContext }, { "database", string.IsNullOrEmpty(database) ? null : database } };
+        }
+
+        public override async Task<IReadOnlyDictionary<string, object>> GetRoutingTable(IConnection connection, string database, string impersonatedUser, Bookmark bookmark)
```
AndyHeap-NeoTech, line 160: What was the reason to override this? I can't see any obvious difference from the parent class implementation.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/RoutingTableManager.cs
```diff
@@ -258,31 +259,27 @@ private Task PrependRoutersAsync(IRoutingTable routingTable, ISet<Uri> uris)
...
+                }
                 catch (SecurityException e)
                 {
                     _logger?.Error(e,
                         "Failed to update routing table from server '{0}' for database '{1}' because of a security exception.",
                         router, database);
                     throw;
                 }
-                catch (FatalDiscoveryException e)
+                catch (ClientException e)
```
AndyHeap-NeoTech, line 273: Nice tidying up by using the parent exception.

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV4_0.cs
thelonelyvulpes: in v3 we pass null instead of bookmark in RunInAutoCommitTransactionAsync then in v4+ we pass the bookmark

AndyHeap-NeoTech (approved):
All good.

PR #605, opened 2022-04-08
Introduce GetServerInfoAsync to IDriver, unify behaviour of VerifyConnectivityAsync

thelonelyvulpes:
- Update VerifyConnectivityAsync on driver to be inline with other drivers.
- Introduce GetServerInfoAsync to driver, which provides connection validation with server details.
- Implement Testkit flags for aforementioned functions


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -378,12 +378,22 @@ public Task CloseAsync()
...
-        
-        public async Task VerifyConnectivityAsync()
+
+        public async Task<IServerInfo> VerifyConnectivityAndGetInfoAsync()
         {
-            // Establish a connection with the server and immediately close it.
-            var connection = await AcquireAsync(Simple.Mode, Simple.Database, null, Simple.Bookmark).ConfigureAwait(false);
-            await connection.CloseAsync().ConfigureAwait(false);
+            var connection = await AcquireAsync(AccessMode.Read, null, null, null)
+                .ConfigureAwait(false) as IPooledConnection;
```
AndyHeap-NeoTech, line 385: Is there a benefit to using the _as_ keyword with _var_ rather than typing the variable? e.g. _IPooledConnection connection =_ ...

I genuinely don't know. Other than stylistic choice.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -128,22 +126,28 @@ public Task CloseAsync()
...
             return Task.CompletedTask;
         }
-
-        public async Task VerifyConnectivityAsync()
+        
+        public async Task<IServerInfo> VerifyConnectivityAndGetInfoAsync()
         {
-            // As long as there is a fresh routing table, we consider we can route to these servers.
-            try
+            var supportsMultiDb = await SupportsMultiDbAsync().ConfigureAwait(false);
```
AndyHeap-NeoTech: Would this be better inside the try/catch block? Or do we want any potential errors from this to propagate as them selves rather than being wrapped in the VerifyConnectivity errors?

AndyHeap-NeoTech (approved):

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
thelonelyvulpes: AcquireAsync returns an IConnection so it needs to be downcast to allow us to release the connection.

I could use casting instead of as but as is preferred in .net (generally).

thelonelyvulpes: I also prefer to use var for all local variables, so you don't get a stylistic mix of type names and var.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscovery.cs
```diff
@@ -100,51 +89,5 @@ public static Uri BoltRoutingUri(string address)
 
             return builder.Uri;
         }
-
-        private class SingleConnectionBasedConnectionProvider : IConnectionProvider
```
thelonelyvulpes, line 104: no references to this class exist.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/ClusterDiscovery.cs
```diff
@@ -20,72 +20,61 @@
...
     {
-        private readonly ILogger _logger;
-        private readonly IDictionary<string, string> _context;
-
-        private const string GetRoutingTableProcedure = "CALL dbms.cluster.routing.getRoutingTable($context)";
-
-        private const string GetRoutingTableForDatabaseProcedure = "CALL dbms.routing.getRoutingTable($context, $database)";
-
-
-        public ClusterDiscovery(IDictionary<string, string> context, ILogger logger)
```
thelonelyvulpes, line 39: none of this is used, so removed it.

AndyHeap-NeoTech (approved):

PR #613, opened 2022-05-11
Retriable exceptions

thelonelyvulpes:
- Neo4jExceptions return bool if can be retried using CanBeRetried Property.
- Compatability mapping for 4.4 servers and 5.0 driver for codes:
   - Neo.TransientError.Transaction.Terminated
   - Neo.TransientError.Transaction.LockClientStopped
- Rename exception extensions method from .IsRetriableError to .CanBeRetried.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/IO/MessageSerializers/FailureMessageV1Serializer.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/IO/MessageSerializers/FailureMessageV1Serializer.cs
@@ -21,10 +21,27 @@ using static Neo4j.Driver.Internal.Protocol.BoltProtocolV3MessageFormat;
 
 namespace Neo4j.Driver.Internal.IO.MessageSerializers
 {
-    internal class FailureMessageSerializer : ReadOnlySerializer
+    internal class FailureMessageV1Serializer : ReadOnlySerializer
     {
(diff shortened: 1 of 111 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/IO/MessageSerializers/FailureMessageV1Serializer.cs
```diff
@@ -21,10 +21,27 @@
 
 namespace Neo4j.Driver.Internal.IO.MessageSerializers
 {
-    internal class FailureMessageSerializer : ReadOnlySerializer
+    internal class FailureMessageV1Serializer : ReadOnlySerializer
```
AndyHeap-NeoTech: Would it be better to name this FailureMessageV3 serializer to match with the protocol naming conventions.

AndyHeap-NeoTech (commented):
Just one question on the naming of the V1 serializer. The rest of it looks like a good implementation.

Neo4j.Driver/Neo4j.Driver/Internal/IO/MessageSerializers/FailureMessageV1Serializer.cs
thelonelyvulpes: I've kept it in keeping with other message serializers and split them by namespace instead of class names.

AndyHeap-NeoTech (approved):

PR #618, opened 2022-06-01
Remove null handling for Node/Relationship Ids

thelonelyvulpes:
Following update to specification remove handling for null id on relationships and nodes.


AndyHeap-NeoTech (approved):
Makes it nice and simple.

thelonelyvulpes:
~Broken PR build due to whitelist check, re-creating PR.~
Fixed.

PR #630, opened 2022-06-28
[5.0] Send ZonedDateTimes as UTC

thelonelyvulpes:


Neo4j.Driver/Neo4j.Driver/Types/ZoneId.cs
```diff
@@ -38,7 +37,14 @@ internal ZoneId(string id)
 
         internal override int OffsetSecondsAt(DateTime dateTime)
         {
-            return (int) TemporalHelpers.GetTimeZoneInfo(Id).GetUtcOffset(dateTime).TotalSeconds;
+            try
```
thelonelyvulpes: To make the temporal tests pass I had to add this. Feel like this is a bug.

AndyHeap-NeoTech (approved):
Implementation looks good. I'll have to trust that the never ending gotchas of UTC are handled as well as we can :)

PR #631, opened 2022-07-15
[4.4]Make DbHits and Rows optional in plans.

thelonelyvulpes:
When profiling a plan, any procedure calls will not return dbhits/rows therefore it should not be mandatory.


AndyHeap-NeoTech (approved):
Yep, simple change with tests as you described :) Looks good.

PR #632, opened 2022-08-04
Introduce Bookmark manager

thelonelyvulpes:
Adds new bookmark manager and user definable interfaces for bookmark management.


Neo4j.Driver/Neo4j.Driver/Config.cs
```diff
@@ -219,6 +219,6 @@ public long FetchSize
         /// Used to get and set the User Agent string. If not used the default will be "neo4j-dotnet/x.y" 
         /// where x is the major version and y is the minor version.
         /// </summary>
-        public string UserAgent { get; set; } = ConnectionSettings.DefaultUserAgent;
+        public string UserAgent { get; internal set; } = ConnectionSettings.DefaultUserAgent;
```
AndyHeap-NeoTech: Is this 'internal' on other drivers. I have a feeling that this functionality might be used by some customers/ SDN I think has a use case for updating it.

AndyHeap-NeoTech (approved):
Great stuff. Just one question around the setting of Agent string to internal.

PR #633, opened 2022-08-04
[5.0] Make DbHits and Rows optional for plans

thelonelyvulpes:
Profiled plans can omit dbhits & rows, so should be optional.


AndyHeap-NeoTech (approved):
Yep, simple change with tests as you described :) Looks good.

PR #637, opened 2022-09-02
Fix 5.0 Builds

thelonelyvulpes:
- Utc encoded date times would throw when offset was > 14 hours.
- Stress test support for 5.0 clustering.

Merged https://github.com/neo4j/neo4j-dotnet-driver/pull/636 as both required to fix build


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Stress/CausalClusterStressTests.cs
```diff
@@ -146,6 +152,42 @@ private ClusterAddresses DiscoverClusterAddresses()
             return new ClusterAddresses(followers, readReplicas);
         }
 
+        private bool UsingBoltMoreThan5_0()
+        {
+            return _driver.GetServerInfoAsync().GetAwaiter().GetResult().ProtocolVersion
```
AndyHeap-NeoTech, line 157: This makes me think we should have an internal method in the driver to do this :)

AndyHeap-NeoTech (approved):
Looks good and fits with our pairing on it.

PR #640, opened 2022-09-07
Reactive & Simple Transaction lifetimes.

thelonelyvulpes:
Update Rx and Simple driver Tx lifetime changes.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Reactive/Internal/InternalRxSession.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Reactive/Internal/InternalRxSession.cs
@@ -122,6 +122,26 @@ namespace Neo4j.Driver.Internal
             return RunTransaction(AccessMode.Write, work, action);
         }
 
+        public IObservable<T> ExecuteRead<T>(Func<IRxRunnable, IObservable<T>> work)
+        {
+            return ReadTransaction(work, null);
+        }
+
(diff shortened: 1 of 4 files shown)


Neo4j.Driver/Neo4j.Driver.Reactive/Internal/InternalRxSession.cs
```diff
@@ -122,6 +122,26 @@ public IObservable<T> WriteTransaction<T>(Func<IRxTransaction, IObservable<T>> w
             return RunTransaction(AccessMode.Write, work, action);
         }
 
+        public IObservable<T> ExecuteRead<T>(Func<IRxRunnable, IObservable<T>> work)
+        {
+            return ReadTransaction(work, null);
```
AndyHeap-NeoTech, line 127: Just wondering if leaving the internal method name(s) ReadTransaction and WriteTransaction as is makes sense, or does it introduces confusion with the deprecated methods?

AndyHeap-NeoTech (approved):
Just one question

Neo4j.Driver/Neo4j.Driver.Reactive/Internal/InternalRxSession.cs
thelonelyvulpes: fair comment, I will avoid using the deprecated methods in the new methods.

PR #644, opened 2022-10-06
Stress test logging

thelonelyvulpes:
Add some environment config so can log details when Stress testing as they can be unpredictable and hard to debug.


AndyHeap-NeoTech (approved):
All good.

PR #649, opened 2022-10-14
Spike auth provider

thelonelyvulpes:


thelonelyvulpes:
Only opened to share a spike at solving an issue.

PR #650, opened 2022-10-20
Query notification filters

thelonelyvulpes:
Implementation of new cypher notifications.
- Adds client filters for cypher notifications.
- Adds Bolt 5.1


Neo4j.Driver/Neo4j.Driver/Internal/Messaging/V3/BeginMessage.cs
```diff
@@ -17,7 +17,6 @@
 
 using System;
 using System.Collections.Generic;
-using Neo4j.Driver;
```
thelonelyvulpes, line 20: undo

thelonelyvulpes:
The API will be reconsidered.

PR #652, opened 2022-10-20
Update example code.

thelonelyvulpes:


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
```diff
@@ -21,1127 +21,1077 @@
...
+            using var session = Driver.Session();
+            var persons = session.ExecuteRead(tx => tx.Run("MATCH (a:Person) RETURN a.name AS name").ToList());
+            return persons.Sum(person => session.ExecuteWrite(tx =>
+            {
+                var result = tx.Run("MATCH (emp:Person {name: $person_name}) " +
+                                    "MERGE (com:Company {name: $company_name}) " +
+                                    "MERGE (emp)-[:WORKS_FOR]->(com)",
+                    new {person_name = person["name"].As<string>(), company_name = companyName});
+                result.Consume();
+                return 1;
```
RichardIrons-neo4j, line 620: probably just being dumb but why does this just return a hardcoded 1?

RichardIrons-neo4j (commented):
All looks fine to me but somebody else should probably check too. I did put one note but it was just a question and doesn't need a change.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
thelonelyvulpes: I don't really know, it was in the example before.

AndyHeap-NeoTech: Looks like it has always been that way. I assume the use of the method doesn't make use of the return value?

thelonelyvulpes: it's used on line 632 to validate this was called twice.

RichardIrons-neo4j: Haha. Thanks for the answer!

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
```diff
@@ -21,1127 +21,1077 @@
...
-                // Given
-                using (var driver = CreateDriverWithCustomizedSecurityStrategy(Uri, User, Password))
                 using (var session = driver.Session())
                 {
-                    // When & Then
-                    session.Run("RETURN 1").Single()[0].As<int>().Should().Be(1);
+                    session.Run("CREATE (a:Person {name: $name})", new {name});
                 }
             }
         }
```
robsdedude, line 261: This code is not being exercised, is it?

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
```diff
@@ -21,1127 +21,1077 @@
...
+            ~DatabaseSelectionExample() => Dispose(false);
+
+            public DatabaseSelectionExample(string uri, string user, string password)
+            {
+                _driver = GraphDatabase.Driver(uri, AuthTokens.Basic(user, password));
             }
 
-            [RequireEnterpriseEdition("5.0.0", VersionComparison.GreaterThanOrEqualTo)]
-            public async void TestUseAnotherDatabaseExampleAsync()
+            public void UseAnotherDatabaseExample()
```
robsdedude, line 869: Is this example being tested somewhere?

robsdedude (approved):
:sweat:

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
thelonelyvulpes: Testkit runs them as part of the integration test suite.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Examples.cs
thelonelyvulpes: yup(ln:821, 845)

PR #654, opened 2022-10-27
Expose bookmark manager config on session

thelonelyvulpes:
Fix bookmark manager not being configurable on session.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
@@ -15,19 +15,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Neo4j.Driver.Internal;
-
 namespace Neo4j.Driver.Experimental;
 
 /// <summary>
-/// Methods being considered for moving to the Neo4j.Driver.GraphDatabase.
+/// Methods being considered for moving to the Neo4j.Driver.
+/// There is no guarantee that anything in Neo4j.Driver.Experimental namespace will be in a next minor version.
 /// </summary>
 public static class ExperimentalExtensions
 {
     /// <summary>
-    /// Experimental: Bookmark Manager API is still under consideration. <br/>
-    /// Gets a new <see cref="IBookmarkManagerFactory"/>, which can construct a default implementation of <see cref="IBookmarkManager"/>.<br/>
-    /// <see cref="IBookmarkManager"/> instances can be passed to <see cref="SessionConfigBuilder"/> when opening a new session.
+    /// There is no guarantee that anything in Neo4j.Driver.Experimental namespace will be in a next minor version.
+    /// Sets the <see cref="IBookmarkManager"/> for maintaining bookmarks for the lifetime of a session.
     /// </summary>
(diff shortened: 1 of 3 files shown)


Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
```diff
@@ -15,19 +15,23 @@
...
+    /// There is no guarantee that anything in Neo4j.Driver.Experimental namespace will be in a next minor version.
+    /// Sets the <see cref="IBookmarkManager"/> for maintaining bookmarks for the lifetime of a session.
     /// </summary>
-    public static IBookmarkManagerFactory BookmarkManagerFactory => new BookmarkManagerFactory();
+    /// <param name="builder">this builder.</param>
+    /// <param name="bookmarkManager">Instance of <see cref="IBookmarkManager"/>.</param>
+    /// <returns>this <see cref="SessionConfigBuilder"/> instance.</returns>
+    public static SessionConfigBuilder WithBookmarkManager(this SessionConfigBuilder builder, IBookmarkManager bookmarkManager)
+    {
+        return builder.WithBookmarkManager(bookmarkManager);
```
RichardIrons-neo4j, line 35: I don't get what's going on here, sorry - this looks like if the extension method wasn't there it would just be the same code?

Neo4j.Driver/Neo4j.Driver/SessionConfig.cs
```diff
@@ -116,7 +116,7 @@ public IEnumerable<Bookmarks> Bookmarks
 
         /// <summary>
         /// Allows the specification of a username that the user wants to impersonate for the duration of the
-        /// session. Once set this cannot be changed for the duration of the sessions liftime. 
+        /// session. Once set this cannot be changed for the duration of the sessions lifetime.
```
RichardIrons-neo4j: Should be `session's lifetime`.

Neo4j.Driver/Neo4j.Driver/SessionConfig.cs
```diff
@@ -218,7 +218,7 @@ public SessionConfigBuilder WithFetchSize(long size)
...
```
RichardIrons-neo4j: Apostrophe again

RichardIrons-neo4j (commented):
Just a question (again) and some missing apostrophes.

Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
thelonelyvulpes: the experimental namespace means it isn't committed to be non-breaking, and as we might need to change the API, the code in SessionConfigBuilder is internal.
the experimental API lets users opt-in to using it only if they use the experimental namespace.
As the bookmark manager has a known issue when using `use dbname` currently. I didn't want to commit to having it fully supported until that is resolved.

RichardIrons-neo4j: Ah great, thanks! Yeah makes perfect sense.

RichardIrons-neo4j (approved):
All looks good.

PR #656, opened 2022-10-31
Simplify creation of bookmark manager with default config

RichardIrons-neo4j:


AndyHeap-NeoTech (approved):
All present and correct.

Neo4j.Driver/Neo4j.Driver/Internal/BookmarkManagerFactory.cs
```diff
@@ -15,13 +15,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
```
thelonelyvulpes: don't think this is needed.

thelonelyvulpes (approved):

PR #657, opened 2022-11-17
Restores LastBookmark on Session in Neo4j.Driver.Simple.

thelonelyvulpes:
When replacing LastBookmark with LastBookmarks we accidentally removed the old property not following the planned deprecation policy, as such it is restored and marked as obsolete and due for removal in 6.0 of the driver.


AndyHeap-NeoTech (approved):
LGTM

PR #663, opened 2022-12-09
Refactor of bolt protocol, connections, messages, serializers.

thelonelyvulpes:
We have refactored bolt protocol and low level connection logic.
Added an editor config.


AndyHeap-NeoTech (approved):
50k lines. Bam! Throw it in there.

Issue #664, opened 2022-12-14
[Feature Request] Ability to refresh auth token


morningpuddle:
To help us understand your issue, please specify important details, primarily:

- Neo4j version: AWS Neptune
- Neo4j Mode: N/A
- Driver version: All
- Operating system: Linux

We are using OpenCypher with AWS Neptune as the endpoint, with Neo4j DotNet Driver with bolt protocol. We generate AWS V4 signature and pass it in as a basic auth to the driver. The request is signed using temporary credentials which expires in 5 minutes. 

- **Steps to reproduce**

With following 3 files (attached with .txt rather than .cs extension, since github doesn't allow uploading .cs files):
 * [HelloWorldExample.txt](https://github.com/neo4j/neo4j-dotnet-driver/files/10231294/HelloWorldExample.txt)
 * [NeptuneAuthToken.txt](https://github.com/neo4j/neo4j-dotnet-driver/files/10231296/NeptuneAuthToken.txt)
 * [Program.txt](https://github.com/neo4j/neo4j-dotnet-driver/files/10231297/Program.txt)

and dependencies 
```
Neo4j.Driver=4.3.0
AWSSDK.Core=3.7.102.1
```

I want to be able to mint a new signature when the new connection is being created after lifetime has been reached (eg. after 30 min for below case; I believe it's [1 hour by default](https://github.com/neo4j/neo4j-dotnet-driver/blob/ad2c0dc4c70cbdc8830562369c93d32d424f0158/Neo4j.Driver/Neo4j.Driver/Config.cs#L161) when not specified), such that not the the old expired signature is used.
```
# in HelloWorldExample file
        public HelloWorldExample()
        {
            var authToken = new NeptuneAuthToken(AccessKey, SecretKey, Region).GetAuthToken(Host);
            _driver = GraphDatabase.Driver(Url, authToken, o => o.WithMaxConnectionLifetime(TimeSpan.FromMinutes(30)).WithEncryptionLevel(EncryptionLevel.Encrypted));
        }

```

- Actual behavior

The first generated signature is getting used for authentication of the new connection, such that it fails with:
```
# then fails 
< 
Unhandled exception. Neo4j.Driver.DatabaseException: "Unexpected server exception 'Signature expired: 20221214T182326Z is now earlier than 20221214T182327Z (20221214T182827Z - 5 min.)'"
   at Neo4j.Driver.Internal.Connector.PooledConnection.OnErrorAsync(Exception error)
[...]

PR #667, opened 2022-12-16
Added the ability to refresh credentials in the auth token

vkagamlyk:
Add the ability to Refresh Auth Token Credentials

Currently when credentials expire, GraphDatabase.Driver will continue trying to connect with the outdated login/password. There is no ability to update or refresh login credentials (See https://github.com/neo4j/neo4j-dotnet-driver/issues/664).

This PR introduces the ability to pass an optional tokenRefresher function to an AuthToken object which will change credentials as they become obsolete.

There are no breaking changes introduced in this PR.

There is a test which validates these modifications by using a simple tokenRefresher which increments the credentials and verifies that the auth token is being updated as expected.

Also tested with connection idle timeout 20 seconds and interval between requests 30 seconds, new credentials were correctly used to create a new connection.

Intended use
`var authToken = new SecretsProvider().GetNeo4jToken();
var driver = GraphDatabase.Driver(Url, authToken, o => o.WithEncryptionLevel(EncryptionLevel.Encrypted));`

`GetNeo4jToken` will receive credits from the secret manager and update them if necessary. This is useful because when changing the password, there is no need to restart the microservices that are connected to the database.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/AuthTokenTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/AuthTokenTests.cs
@@ -103,6 +103,36 @@ namespace Neo4j.Driver.Tests
                 nums["Two"].Should().Be(2);
                 nums["Three"].Should().Be(3);
             }
+
+            [Fact]
+            public void ShouldCreateCustomAuthTokenWithRefresher()
+            {
+                var runNumber = 0;
+
+                var authToken = AuthTokens.Custom("zhenli", null, "foo", "custom", null, Refresher);
+                var dict = authToken.AsDictionary();
+
+                dict.Count.Should().Be(4);
+                dict["scheme"].Should().Be("custom");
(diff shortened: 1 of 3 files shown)


vkagamlyk:
I signed CLA

nuno99costa:
Any updates on this? Definitely something I'm looking forward to use.

AndyHeap-NeoTech:
Hi vkagamlyk,

Reiterating our reply on you PR for the python driver. Thank you for the effort and time spent on these PRs. We have ongoing work to implement a refresh mechanism on the drivers in a more general way that should be released Q1 of this year (all things going well that is). 

There are considerations to be made around the interaction of a token refresh, the retry mechanism that the drivers supply and the connection pooling. This is why our solution is more involved. 

As a side note, did you intend to raise the PR against the 4.3 version of the driver? This version is out of support and will not be updated. The current LTS version is 4.4, and the latest driver version is 5.4.

PR #671, opened 2023-01-10
Fix IsRetriable not being available on the public API

thelonelyvulpes:
IsRetriable was supposed to be available as a public call, this resolves it be marked as internal and incorrectly named.


AndyHeap-NeoTech (approved):
All present and correct. Approved.

PR #673, opened 2023-01-29
Switched to AsyncKeyedLock

MarkCiliaVincenti:
Will increase performance and reduce memory allocations.


AndyHeap-NeoTech:
Hi Mark. Thankyou for taking the time to contribute to the .net driver. We are unable to look at the PR until you have agreed to and submitted the CLA. But going from the title of the PR it looks like you may be introducing a new library and dependency. The driver has 0 dependencies by design/choice and we wish to keep it that way at the moment. If you have introduced a dependency we will have to close the PR without accepting it.

MarkCiliaVincenti:
Yes I did add a dependency. Closing then.

PR #675, opened 2023-02-01
Transformation fluent working

RichardIrons-neo4j:
Mainly a refactor of `EagerResult` to `EagerResult<T>` so that it can contain records of any type, not just `IRecord`.

As part of filtering I will refactor a bit more so that `IAsyncEnumerable` is the main thing being worked with, maybe removing `IResultTransformer`.


AndyHeap-NeoTech (approved):
Of these changes I hearby approve.

PR #676, opened 2023-02-01
Expose IASyncEnumerable in callback to allow async linq to be used

RichardIrons-neo4j:


Neo4j.Driver/Neo4j.Driver/Experimental/FluentQueries/ExecutableQuery.cs
```diff
@@ -23,55 +23,77 @@
...
+        await foreach (var item in enumerable)
+        {
+            result.Add(item);
+        }
+
+        return result;
     }
 
-    public IExecutableQuery WithConfig(QueryConfig config)
+    public IExecutableQuery<T> WithConfig(QueryConfig config)
```
AndyHeap-NeoTech, line 61: Did we make a decision on whether to move to a set of 'With' methods rather than a config object? Something we can add after release if we think it is a better way of doing it.

Neo4j.Driver/Neo4j.Driver/Experimental/EagerResult.cs
```diff
@@ -24,41 +24,54 @@ namespace Neo4j.Driver.Experimental;
 /// There is no guarantee that anything in Neo4j.Driver.Experimental namespace will be in a next minor version.
 /// Complete result from a cypher query.
 /// </summary>
-public sealed class EagerResult : IReadOnlyList<IRecord>
+public sealed class EagerResult<T>
```
AndyHeap-NeoTech, line 24: Not for this line but the licensing text above. Looks like this is using the older text. There is a newer non dated version that should be used. Can be copied from the top of ExecutableQuery.cs

Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
```diff
@@ -15,6 +15,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
```
AndyHeap-NeoTech: Not for this line but the licensing text above. Looks like this is using the older text. There is a newer non dated version that should be used. Can be copied from the top of ExecutableQuery.cs

AndyHeap-NeoTech (approved):

Neo4j.Driver/Neo4j.Driver/Experimental/FluentQueries/ExecutableQuery.cs
RichardIrons-neo4j: Yeah, the config object is spread around quite a lot at the moment, but yes, we'll do it with "withers".

Neo4j.Driver/Neo4j.Driver/Experimental/EagerResult.cs
RichardIrons-neo4j: Done.

Neo4j.Driver/Neo4j.Driver/Experimental/ExperimentalExtensions.cs
RichardIrons-neo4j: Done.

PR #683, opened 2023-02-15
Comply to the latest revision of the ExecuteQuery ADR

fbiville:
Rename stream processor to result transformer

Expose ExecuteQuery's default bookmark manager


fbiville:
Note: some of the renames may be completely inappropriate. Shout at me if that's the case.

AndyHeap-NeoTech:
Just to note: Some of these naming have a bit of flexibility on a per language basis. We are not enforcing a uniform naming across drivers if it really clashes with the idiomacy of things.

RichardIrons-neo4j (approved):
👍

PR #684, opened 2023-02-15
Example change

thelonelyvulpes:
* Create an example tag
* Make an example without finalizers


AndyHeap-NeoTech (approved):
A destructor! How has that gone unnoticed for so long.

PR #693, opened 2023-03-17
Add TryVerifyConnectivity

### LINKED ISSUE #690: [Feature Request] None throwing connectivity-check
I have a case where I do not care for the reason of connectivity issues. I just need to know whether or not i am able to connect, that's it.
Maybe another option, as shown below could be a plausible option?

```csharp
namespace Neo4j.Driver;

public interface IDriver
{
    public Task<bool> TryVerifyConnectivityAsync();
}
```

thelonelyvulpes:


AndyHeap-NeoTech (approved):
Looks like a nice addition to the driver surface.

PR #699, opened 2023-04-20
Refactor Try out pipeline reader

thelonelyvulpes:
Implement [PipeReader](https://learn.microsoft.com/en-us/dotnet/api/system.io.pipelines.pipereader?view=dotnet-plat-ext-8.0) for message reading.
- Pipeline reader allows us to easily read large messages and reduce allocations.
- is Performant
- Packstream serialization over Spans and Sequences
- Message reading with span and sequences ensures safety of the stream as they do not mutate external state.


Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamReader.cs
```diff
@@ -19,14 +19,40 @@
 using System.Collections.Generic;
 using System.IO;
 using Neo4j.Driver.Internal.Connector;
+using Neo4j.Driver.Internal.Messaging;
 
 namespace Neo4j.Driver.Internal.IO;
 
-internal sealed class PackStreamReader
+internal interface IPackStreamReader
```
thelonelyvulpes, line 26: Need to undo, there is no use for the interface as the classes for Span and Sequence Readers are ref structs

PR #707, opened 2023-05-30
Update README.md

thelonelyvulpes:
Remove execute query dicussion link from readme


thelonelyvulpes:
the discussion is now closed so we can remove the link from the readme

PR #722, opened 2023-07-18
SPIKE - Telemetry collection with activitysource

RichardIrons-neo4j:
Spike to do telemtry collection with built-in stuff from .NET instead.

There's some code in there to only record the outermost type of API call, since the "unmanaged transaction" type is called from within other APIs.


Neo4j.Driver/Neo4j.Driver/Internal/AsyncSession.cs
```diff
@@ -248,7 +250,14 @@ public Task ExecuteReadAsync(Func<IAsyncQueryRunner, Task> work, Action<Transact
         Func<IAsyncQueryRunner, Task> work,
         Action<TransactionConfigBuilder> action = null)
     {
-        return RunTransactionAsync(AccessMode.Write, work, action);
+        // need this so the method has a return type
+        async Task<byte> WorkProxy(IAsyncQueryRunner qr)
+        {
+            await work(qr);
```
thelonelyvulpes: configure await

Neo4j.Driver/Neo4j.Driver/Internal/AsyncSession.cs
```diff
@@ -302,6 +311,12 @@ private async Task<Bookmarks> GetBookmarksAsync()
         AccessMode mode,
         Func<IAsyncTransaction, Task<T>> work,
         Action<TransactionConfigBuilder> action)
+    {
+        using var _ = Driver.TelemetryManager.StartApiActivity(QueryApiType.TransactionFunction);
+        return RunTransactionImplAsync(mode, work, action);
```
thelonelyvulpes: if RunTransactionImplAsync awaits before reading the activity it will be unset so best to await here

Neo4j.Driver/Neo4j.Driver/Internal/AsyncTransaction.cs
```diff
@@ -84,6 +85,8 @@ public void UpdateBookmarks(Bookmarks bookmarks, IDatabaseInfo dbInfo = null)
 
     public override Task<IResultCursor> RunAsync(Query query)
     {
+        using var _ = Driver.TelemetryManager.StartApiActivity(QueryApiType.UnmanagedTransaction);
```
thelonelyvulpes: We shouldn't need it on run, we can read this on the begin transaction right?

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -224,6 +232,30 @@ public async Task SendAsync()
...
+        if (Version < BoltProtocolVersion.V5_4)
+        {
+            // telemetry not supported before 5.4
+            return;
+        }
+
+        lock (_telemetryCollector)
+        {
+            // add a telemetry message if we have enough now
+            if (_messages.Any(m => m is RunWithMetadataMessage) && _telemetryCollector.BatchSizeReached)
```
thelonelyvulpes, line 253: Should probably be TransactionStartingMessage that covers all Tx APIs and session.run

RichardIrons-neo4j: I think talk about this when the spike becomes reality (I agree though).

PR #724, opened 2023-08-14
Implement pipelined begin for executable query

thelonelyvulpes:
Optimizes running executable queries by removing a server round trip.


AndyHeap-NeoTech (commented):
Looks good, pretty straightforward. Again we find ourselves passing parameters down the stack. Was going to question the switch to using Session instead of AsyncSession, then realised it is just a wrapper method for Session.

thelonelyvulpes:
Awaiting: https://github.com/neo4j-drivers/testkit/pull/588

thelonelyvulpes:
> Looks good, pretty straightforward. Again we find ourselves passing parameters down the stack. Was going to question the switch to using Session instead of AsyncSession, then realised it is just a wrapper method for Session.

I have moved the parameter into an object, going forward we can push values on to that instead of adding a parameter to several methods. There's still a couple more bolt protocol methods that could do with a parameter object treatment but I think tackling them one by one keeps the size of changes down.

AndyHeap-NeoTech (approved):
LGTM

PR #725, opened 2023-08-15
ADR 019 password rotation

RichardIrons-neo4j:
This PR updates the preview feature "re-auth" significantly. The changes allow for catering to a wider range of use cases including simple password rotation.

As part of this PR, all auth-related namespaces have been moved to preview - previously some did not have this, although the classes therein would not have been usable.

Since this is a preview feature all changes here are breaking changes.

The `OnTokenExpiredAsync` method in the `IAuthTokenManager` interface was removed, and a new `HandleSecurityExceptionAsync` method was added in its place.

The `ExpirationBased` method in `AuthTokenManagers` was renamed to `Bearer`, and a new `Basic` method was added to cater for password rotation.


Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -373,10 +375,12 @@ public async Task ValidateCredsAsync()
             token = await AuthTokenManager.GetTokenAsync().ConfigureAwait(false);
         }
 
-        // The token has changed or the connection needs to re-authenticate but can use the same credentials.
-        if (AuthorizationStatus == AuthorizationStatus.AuthorizationExpired || !token.Equals(AuthToken))
+        var authExpired = AuthorizationStatus == AuthorizationStatus.AuthorizationExpired;
+        if (authExpired || !token.Equals(AuthToken))
```
AndyHeap-NeoTech, line 387: Is there a situation here where the code would need to handle a null token being supplied from the AuthTokenManager. Previous code threw an exception in that case (see line 354 for conditional on null check).

AndyHeap-NeoTech (approved):
I had one question about handling a potential null token. All looks good to me with nothing obvious standing out.

Neo4j.Driver/Neo4j.Driver/Neo4j.Driver.csproj
```diff
@@ -39,21 +39,21 @@
         <WarningLevel>2</WarningLevel>
     </PropertyGroup>
     <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
-        <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0"/>
```
thelonelyvulpes, line 42: for whatever reason our configs are not aligned on what this xml should have, we need to pick a style.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -299,19 +299,13 @@ public void OnPoolMemberException(IPooledConnection connection, Exception except
...
-                }
-            }
+            connection.AuthorizationStatus = AuthorizationStatus.SecurityError;
         }
 
         if (exception is AuthorizationException)
         {
+            // if an auth exception occured the pool member, all connections in the pool that are using that token
+            // should be closed. This is because the token is now invalid and all connections using it will fail.
             foreach (var conn in _inUseConnections)
```
thelonelyvulpes: do we care about setting it for idle connections, _inUseConnections are connections that another session or transaction currently has borrowed.

thelonelyvulpes (approved):

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
RichardIrons-neo4j: I _think_ it's right to let those connections fail the next time they're attempted to be used?

thelonelyvulpes: why do we want idle connections to lazily discover the auth token is expired but set in use connections as expired?

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
RichardIrons-neo4j: Added a null check.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
RichardIrons-neo4j: Yeah, thinking about it, you're right, let me alter that.

robsdedude (commented):
I'm a bit late to the party, but please note that the TestKit tests for this feature introduced a new feature flag `API_RETRYABLE_EXCEPTION = "Feature:API:RetryableExceptions"`. For better test coverage, I highly recommend implementing it in the backend.

PR #726, opened 2023-08-16
Optimize record creation

thelonelyvulpes:


Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
```diff
@@ -21,24 +21,38 @@ namespace Neo4j.Driver.Internal.Result;
 
 internal class Record : IRecord
 {
-    public Record(string[] keys, object[] values)
-    {
-        Throw.ProtocolException.IfNotEqual(keys.Length, values.Length, nameof(keys), nameof(values));
-
-        var valueKeys = new Dictionary<string, object>();
```
thelonelyvulpes, line 28: This could have been allocated with a correct initial length.

Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
```diff
@@ -21,24 +21,38 @@ namespace Neo4j.Driver.Internal.Result;
...
-    public object this[int index] => Values[Keys[index]];
-    public object this[string key] => Values[key];
+    public object this[int index] => _values[index];
+    public object this[string key] => _values[_fieldIndexes[key]];
+
+    public IReadOnlyDictionary<string, object> Values
+    {
+        get
+        {
+            var valueKeys = new Dictionary<string, object>(_values.Length);
```
thelonelyvulpes, line 47: we are building this once per call, it could be stored for reuse for the user code.
```csharp
var a = record.Values["a"];
var b = record.Values["b"];
```

PR #730, opened 2023-09-20
Refactor the sprawl of driver and connection configs

thelonelyvulpes:
Refactor of driver config and its usage across the driver.
- Driver level Config is no longer duplicated or lazily, validation should be completed as early as possible.
- Remove null-conditional member access for logging
  - Previously the only way to get a null logger was to define the logger as null explictly in the driver, which would replace the no-op logger at `NullLogger.Instance` instead we check if it is null there and ignore the null parameter. half of the code also then wraps the logger in a prefix logger so would end up not being null, but still no-oping


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Stress/StressTest.cs
```diff
@@ -829,39 +829,6 @@ public bool CompareAndSet(T value, T comparand)
         }
     }
 
-    private class MonitoredPooledConnectionFactory : IPooledConnectionFactory
```
thelonelyvulpes, line 832: unused

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -144,4 +144,15 @@ private bool HasBeenIdleForTooLong(IPooledConnection connection)
...
+
+
+    public static bool IsTimeoutDetectionEnabled(TimeSpan timeout)
+    {
+        return timeout.TotalMilliseconds >= 0;
+    }
+
+    public static bool IsTimeoutDetectionDisabled(TimeSpan timeout)
+    {
+        return timeout.TotalMilliseconds < 0;
```
RichardIrons-neo4j: is a second function needed? And if it is, it would be better to make Disabled => !Enabled so that the logic is only encoded into one place in the code.

RichardIrons-neo4j (approved):
Seems fine, only one minor comment.

AndyHeap-NeoTech (commented):
Looks good and nothing obviously out of place. The removal of repetitive member initialisation in a load of the constructors brings me joy :)

PR #737, opened 2023-10-10
Transaction error handling

thelonelyvulpes:
This change introduces a mechanism to remove undefined behaviors when the user code catches and ignores an exception and attempts to use a terminated transaction further.
- When a result is a constituent of the transaction, and that transaction is terminated, all results are invalidated.


thelonelyvulpes:
Requires https://github.com/neo4j-drivers/testkit/pull/601 & https://github.com/neo4j-drivers/testkit/pull/603

Neo4j.Driver/Neo4j.Driver/Internal/AsyncTransaction.cs
```diff
@@ -108,7 +136,10 @@ public async Task RollbackAsync()
     {
         try
         {
-            await DiscardUnconsumed().ConfigureAwait(false);
+            if (TransactionError == null)
+            {
+                await DiscardUnconsumed().ConfigureAwait(false);
```
thelonelyvulpes: We should scrutinize this check

Neo4j.Driver/Neo4j.Driver/Internal/Result/ConsumableResultCursor.cs
```diff
@@ -48,19 +52,31 @@ public Task<string[]> KeysAsync()
...
+        if (Transaction.IsErrored(out var error))
+        {
+            throw new TransactionTerminatedException(error);
+        }
+    }
+
     public Task<IRecord> PeekAsync()
     {
         AssertNotConsumed();
+        AssertTransactionValid();
```
thelonelyvulpes: Doing this, each record seems suboptimal, so if we can think of a better place, we should try. similarly for fetch

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/Utility/FakeTransaction.cs
```diff
@@ -0,0 +1,90 @@
...
+            "Something very wrong has happened and an illegal function has been called.");
+    }
+
+    public Task RollbackAsync()
+    {
+        throw new InvalidOperationException(
+            "Something very wrong has happened and an illegal function has been called.");
+    }
+
+    public bool IsErrored(out Exception ex)
```
thelonelyvulpes: Maybe just null-checking the exception would be better and easier to mock.

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/IBoltProtocol.cs
```diff
@@ -51,7 +51,8 @@ internal interface IBoltProtocol
         IConnection connection,
         Query query,
         bool reactive,
-        long fetchSize = Config.Infinite);
+        long fetchSize,
```
thelonelyvulpes, line 54: We always provide a fetchsize so the overload is redundant.

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolV3.cs
```diff
@@ -199,8 +201,9 @@ public async Task BeginTransactionAsync(IConnection connection, BeginTransaction
             null,
             null,
             null,
-            Config.Infinite,
-            false);
+            Config.Infinite, // Bolt Protocol 3 uses PULLALL so fetch size is ignored.
```
thelonelyvulpes, line 204: I saw this and had a mild panic then realized why it was infinite so I added this for posterity.

Neo4j.Driver/Neo4j.Driver.Tests/Internal/Protocol/BoltProtocolTests.cs
```diff
@@ -153,7 +152,7 @@ public class GetRoutingTableAsyncTests
             public async Task ShouldThrowAnExceptionWhneNullConnection()
```
RichardIrons-neo4j:
```suggestion
            public async Task ShouldThrowAnExceptionWhenNullConnection()
```

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/Utility/FakeTransaction.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Neo4j.Driver.Internal;
+
+/// <summary>
+/// Supports cursors on auto commit functions so we don't need to null check and check error.
+/// this will always return false for the transaction has errored because the transaction doen't exist.
+/// </summary>
+internal sealed class FakeTransaction : IInternalAsyncTransaction
```
RichardIrons-neo4j: Could there be a better name for this class? Fake sounds a bit like it's something to do with testing.

RichardIrons-neo4j (approved):

PR #740, opened 2023-10-16
Tidying and minor refactoring

RichardIrons-neo4j:
Some minor tidying and refactoring of the mapping code.


thelonelyvulpes (approved):

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/BuiltMapper.cs
```diff
@@ -56,18 +56,18 @@ public void AddWholeObjectMapping(Func<IRecord, TObject> mappingFunction)
...
         var propertyType = propertySetter.GetParameters()[0].ParameterType;
         var asMethod = _asGenericMethod.MakeGenericMethod(propertyType);
-        var getter = _mappingSourceDelegateBuilder.GetMappingDelegate(mappingSource);
+        var getter = _mappingSourceDelegateBuilder.GetMappingDelegate(entityMappingInfo);
         AddMapping(propertySetter, GetValue);
 
         object GetValue(IRecord record)
         {
-            var found = getter(record, out var value);
+            getter(record, out var value);
```
thelonelyvulpes: Why aren't we checking if it is found anymore?

thelonelyvulpes:
```suggestion
            var _ = getter(record, out var value);
```
We could make the unused parameter more explicit

RichardIrons-neo4j: I put in a better bit so that `found` is handled separately to getting a null value out (since if `found` is false we shouldn't assume anything about `value`).

PR #754, opened 2023-12-04
Introduce TransactionConfig parameter to query config

thelonelyvulpes:
Exposes Transaction config for driver queries.
- Makes TransactionConfig constructor public.
- transactionConfig added to QueryConfig

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
+++ b/Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
@@ -16,55 +16,44 @@
 using System;
 using System.Collections.Generic;
 using Neo4j.Driver.Internal;
-using Neo4j.Driver.Internal.IO;
 
 namespace Neo4j.Driver;
 
 /// <summary>
 /// Configuration object containing settings for explicit and auto-commit transactions. Leave the fields unmodified to use
 /// server side transaction configurations.
-/// <para/>
-/// For example, the following code starts a transaction using server default transaction configurations.
-/// <code>
-/// session.BeginTransaction(b=>{});
-/// </code>
 /// </summary>
 public sealed class TransactionConfig
 {
     internal static readonly TransactionConfig Default = new();
-    private IDictionary<string, object> _metadata;
-    private TimeSpan? _timeout;
-
-    internal TransactionConfig()
-    {
-        _timeout = null;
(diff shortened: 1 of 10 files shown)


Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
```diff
@@ -16,63 +16,61 @@
...
+    /// zero.
     /// </exception>
     public TimeSpan? Timeout
     {
         get => _timeout;
-        internal set
+        init
         {
-            if ((value ?? TimeSpan.Zero) < TimeSpan.Zero)
+            if (!value.HasValue || value.Value == TimeSpan.MaxValue)
```
AndyHeap-NeoTech, line 52: Is this correct? If value doesn't have a value you are still using it to set _timeout...?

Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
```diff
@@ -115,42 +113,47 @@ public sealed class TransactionConfigBuilder
...
     public TransactionConfigBuilder WithTimeout(TimeSpan? timeout)
     {
-        _config.Timeout = FixSubmilliseconds(timeout);
-        return this;
-    }
-
-    private TimeSpan? FixSubmilliseconds(TimeSpan? timeout)
-    {
-        if(timeout == null || timeout == TimeSpan.MaxValue)
+        if (!timeout.HasValue || timeout.Value == TimeSpan.MaxValue)
```
AndyHeap-NeoTech, line 132: Same question as above.

AndyHeap-NeoTech (approved):
Just a couple of questions. Other than that looks good.

AndyHeap-NeoTech (approved):
Just a couple of questions. Other than that looks good.

Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
thelonelyvulpes: that is the current behaviour, it is setting null to null so slightly redundant but it meant slightly fewer lines of code.

Neo4j.Driver/Neo4j.Driver/TransactionConfig.cs
thelonelyvulpes: Same answer as above.

PR #756, opened 2023-12-05
Liveness check

thelonelyvulpes:
This change introduces a liveness check timeout.
- Liveness checks will be completed when borrowing a pooled connection, given that the connection has been idle longer than the timeout.
- The driver, by default, will not add liveness checks, but this can be configured with `WithConnectionLivenessCheckTimeout` on the `ConfigBuilder`.

```csharp
using var driver = GraphDatabase.Driver(uri, creds, 
    cb => cb.WithConnectionLivenessCheckTimeout(TimeSpan.FromSeconds(30));
```

Using a timeout of `0`, all connections will complete a liveness check when taken from the pool; this ensures connections are healthy but comes at the cost of performance when running queries or starting transactions.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Time/FakeTime.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Time/FakeTime.cs
@@ -93,9 +93,9 @@ internal class FakeTime : IDateTimeProvider
 
     public ITimer NewTimer()
     {
-        var t= new FakeTimer();
-        Timers.Add(t);
-        return t;
+        var fakeTimer = new FakeTimer();
+        Timers.Add(fakeTimer);
+        return fakeTimer;
     }
 
     public void Freeze()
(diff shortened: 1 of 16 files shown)


Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Time/FakeTime.cs
```diff
@@ -96,10 +106,32 @@ public void Freeze()
     public void Advance(int milliseconds)
     {
         _frozenTime = Now().AddMilliseconds(milliseconds);
+        foreach (var t in Timers)
+        {
+            t.Advance(milliseconds);
+        }
     }
 
     public void Unfreeze()
```
RichardIrons-neo4j: The only thing I can see here is that freezing and unfreezing will have no effect on the timers, since they are always frozen. I like this solution to the timer problem, but perhaps it could be done with a stopwatch that gets started and stopped, in case testkit ever does something without freezing the time and expects the timers to tick?

(I know this is unlikely but if it did ever happen I think it might cause weird stuff that's hard to track down)

RichardIrons-neo4j (approved):

PR #758, opened 2023-12-13
Protect default memory pool for bolt

thelonelyvulpes:
This change introduces and defaults the pool used for bolt to a hermetic pool, and removes the ability to disable new pipeline reader
- The best pool to use is the shared pool as it means any buffers allocated are shared to the application, but this exposes users to a level of risk, while we should suggest users use the shared pool, the default should remain isolated.


RichardIrons-neo4j (approved):
Looks fine

AndyHeap-NeoTech (approved):
All good. Makes sense to use a separate memory pool as the default.

AndyHeap-NeoTech (approved):

RichardIrons-neo4j (approved):

PR #763, opened 2024-01-02
Fix liveness checks only being enabled with idle timeout

thelonelyvulpes:
fix the liveness check if idle timeout is disabled.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
@@ -37,6 +37,7 @@ internal class ConnectionValidator : IConnectionValidator
     private readonly long _connIdleTimeout;
     private readonly long _livenessTimeout;
     private readonly long _maxConnLifetime;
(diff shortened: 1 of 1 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionValidator.cs
```diff
@@ -48,6 +49,8 @@ internal class ConnectionValidator : IConnectionValidator
         _livenessTimeout = livenessCheckTimeout.HasValue
             ? (long)livenessCheckTimeout.Value.TotalMilliseconds
             : long.MaxValue;
+
+        _idleTimerActive = _connIdleTimeout < long.MaxValue || _livenessTimeout < long.MaxValue;
```
AndyHeap-NeoTech, line 53: You're changes make sense, but I'm wondering about the initial logic that was implemented that you've built on top of. Without more context it looks like it is considering a timeout active if it is set to zero?

AndyHeap-NeoTech (approved):

PR #777, opened 2024-02-09
Find fields and properties via dictionary first if possible

RichardIrons-neo4j:
This change adds a method to RecordPathFinder that will check if the path supplied can be used directly as keys into the record rather than looping through all the fields in the record checking. If this fails then the old method is still used, however in tests the fast method was used most of the time.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DefaultMapper.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DefaultMapper.cs
@@ -57,10 +57,14 @@ internal static class DefaultMapper
                 continue;
             }
 
+            // lower-case the first letter of the property name to be more likely to
+            // match the exact case of the property name in the record
+            var propertyName = char.ToLowerInvariant(property.Name[0]) + property.Name.Substring(1);
+
             // check if there is a MappingSourceAttribute: if there is, use the specified mapping source;
             // if not, look for a property on the entity with the same name as the property on the object
             var mappingSource = property.GetCustomAttribute<MappingSourceAttribute>()?.EntityMappingInfo ??
-                new EntityMappingInfo(property.Name, EntityMappingSource.Property);
+                new EntityMappingInfo(propertyName, EntityMappingSource.Property);
 
             // don't re-map any fields that were already mapped by the constructor
             if (!usedEntitySources.Contains(mappingSource.Path.ToLowerInvariant()))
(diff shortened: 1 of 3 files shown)


Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DefaultMapper.cs
```diff
@@ -57,10 +57,14 @@ public static IRecordMapper<T> Get<T>()
                 continue;
             }
 
+            // lower-case the first letter of the property name to be more likely to
+            // match the exact case of the property name in the record
+            var propertyName = char.ToLowerInvariant(property.Name[0]) + property.Name.Substring(1);
```
thelonelyvulpes, line 62: This seems odd to me, a defined strategy for handling name matching would be much better to me than us guessing

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/RecordPathFinder.cs
```diff
@@ -75,4 +81,40 @@ public bool TryGetValueByPath(IRecord record, string path, out object value)
...
+
+        // if the path matches a field name, we can return the value directly
+        if (record.Values.TryGetValue(path, out value))
+        {
+            return true;
+        }
+
+        // if there's a dot in the path, we can try to split it and check if the first part
+        // matches a field name and the second part matches a property name
+        var dotIndex = path.IndexOf('.');
```
thelonelyvulpes, line 97: should we throw if there is more than 1 `.` or is this validated elsewhere

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/RecordPathFinder.cs
```diff
@@ -75,4 +81,40 @@ public bool TryGetValueByPath(IRecord record, string path, out object value)
 
         return false;
     }
+
+    private bool TryGetValueFast(IRecord record, string path, out object value)
```
thelonelyvulpes: this function seems like a good candidate for a few tests.

thelonelyvulpes (approved):

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DefaultMapper.cs
thelonelyvulpes: I think this is okay actually for this case, I would prefer the logic came from a config on the driver or something that we will be assuming for the default casing will be camel case.

PR #779, opened 2024-02-12
TestKit backend: serialize server address in summary

robsdedude:


robsdedude:
My bad. This makes no sense to implement before we've unified the address behavior (cf. [Trello card (internal)](https://trello.com/c/WjC7KMiT))

PR #780, opened 2024-02-13
Invariant capable records

RichardIrons-neo4j:
This changes the Record type so that it does not contain a dictionary on every instance, instead only  containing the array of values, and references to two dictionaries shared across the whole cursor: one just a straight Dictionary<string, object>, and the other of the same type but using an InvariantCultureCaseInsensitive string comparer, so that values can be looked up using a case insensitive key, used in the object mapping. 

Added Get<T> and TryGet<T> methods to IRecord and IEntity, and added xxxCaseInsensitive versions of those methods to IRecord.

More tests added, tests modified to create records through `TestRecord.Create`.


Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultCursorBuilder.cs
```diff
@@ -139,7 +144,11 @@ public async ValueTask<IResultSummary> ConsumeAsync()
     public void RunCompleted(long queryId, string[] fields, IResponsePipelineError error)
     {
         _queryId = queryId;
-        _fields = fields;
+
+        _fieldLookup = fields.Select((field, index) => (field, index))
```
thelonelyvulpes: can we do this the old fashioned way, and do it in 1 pass.
```
var invariant = new Dictionary<string, int>(fields.Count, StringComparer.InvariantCultureIgnoreCase);
var exact = new Dictionary<string, int>(fields.Count);
for (var i = 0; i < fields.Count; i++)
{
    invariant.Add(fields[i], i);
    exact.Add(fields[i], i);
}
_invariantFieldLookup = invariant;
_fieldLookup = exact;
```

RichardIrons-neo4j: Done

Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
```diff
@@ -13,34 +13,118 @@
...
+    /// <inheritdoc />
+    IEnumerable<string> IReadOnlyDictionary<string, object>.Keys => Keys;
+
+    /// <inheritdoc />
+    IEnumerable<object> IReadOnlyDictionary<string, object>.Values => _fieldValues;
+
+    /// <inheritdoc />
+    IEnumerator<KeyValuePair<string, object>> IEnumerable<KeyValuePair<string, object>>.GetEnumerator()
+    {
+        return Keys.Select(key => new KeyValuePair<string, object>(key, this[key])).GetEnumerator();
```
thelonelyvulpes:
```suggestion
        return Keys.Select((key, i) => new KeyValuePair<string, object>(key, _fieldValues[i])).GetEnumerator();
```

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultCursorBuilder.cs
```diff
@@ -139,7 +144,20 @@ public async ValueTask<IResultSummary> ConsumeAsync()
     public void RunCompleted(long queryId, string[] fields, IResponsePipelineError error)
     {
         _queryId = queryId;
-        _fields = fields;
+
+        if (fields is not null)
```
thelonelyvulpes, line 148: We probably don't need this check. If we want to include it, we should instead throw a invalid or protocol exception if fields is null

Neo4j.Driver/Neo4j.Driver/Public/IRecord.cs
```diff
@@ -28,11 +28,49 @@ public interface IRecord
     /// <summary>Gets the value specified by the given key.</summary>
     /// <param name="key">The key</param>
     /// <returns>the value specified with the given key.</returns>
-    object this[string key] { get; }
+    new object this[string key] { get; }
```
thelonelyvulpes: we could drop this as it is defined on the IReadOnlyDictionary<string, object> interface and behaves the same.

Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
RichardIrons-neo4j: Done.

Neo4j.Driver/Neo4j.Driver/Public/IRecord.cs
RichardIrons-neo4j: Done.

Neo4j.Driver/Neo4j.Driver/Internal/Result/ResultCursorBuilder.cs
RichardIrons-neo4j: There are a couple of places where null is passed in as the `fields` parameter - `OnFailure` and `OnIgnored`. The old code didn't use to do anything with the array so the check wasn't needed, but now it builds the lookups it needs the check. We could change those places to pass an empty array but then we'd have it creating two empty dictionaries for no reason.

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DictAsRecord.cs
```diff
@@ -27,20 +28,91 @@ internal class DictAsRecord : IRecord
 
     public DictAsRecord(object dict, IRecord record)
```
thelonelyvulpes, line 29: can we make this class sealed

Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
```diff
@@ -13,34 +13,100 @@
...
 // limitations under the License.
 
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Neo4j.Driver.Internal.Result;
 
 internal class Record : IRecord
 {
```
thelonelyvulpes, line 23: let's make this class sealed

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DictAsRecord.cs
```diff
@@ -27,20 +28,91 @@ internal class DictAsRecord : IRecord
...
+            IEntity entity => entity.Properties ?? new Dictionary<string, object>(),
             IReadOnlyDictionary<string, object> dictionary => dictionary,
             _ => throw new InvalidOperationException($"Cannot create a DictAsRecord from a {dict.GetType().Name}")
         };
 
+        // this is only used by the default mapper so make it case insensitive
+        _dict = readOnlyDictionary.ToDictionary(
+            kvp => kvp.Key,
+            kvp => kvp.Value,
+            StringComparer.OrdinalIgnoreCase);
```
thelonelyvulpes: This is ordinal but the other is CultureInvariantIgnoreCase, we should only use one for consistency.

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DictAsRecord.cs
RichardIrons-neo4j: done

Neo4j.Driver/Neo4j.Driver/Internal/Result/Record.cs
RichardIrons-neo4j: done

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DictAsRecord.cs
RichardIrons-neo4j: Changed to CultureInvariantIgnoreCase

thelonelyvulpes (approved):

PR #782, opened 2024-02-28
Change DictAsRecord to have non-surprising case sensitivity; increase…

RichardIrons-neo4j:
Change DictAsRecord to have normal case sensitivity since users may now encounter an instance of the class.

Test coverage for all methods.


thelonelyvulpes (approved):

thelonelyvulpes:
good spot.

PR #784, opened 2024-03-04
move mapping api out of preview namespace

RichardIrons-neo4j:
Move all the mapping classes out of the preview namespace and into `Neo4j.Driver.Mapping`.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Public/Mapping/IMappingBuilder.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Public/Mapping/IMappingBuilder.cs
@@ -16,7 +16,7 @@
 using System;
 using System.Linq.Expressions;
 
-namespace Neo4j.Driver.Preview.Mapping;
+namespace Neo4j.Driver.Mapping;
 
 /// <summary>
 /// Defines a builder for mapping objects from <see cref="IRecord"/>s.
--- a/Neo4j.Driver/Neo4j.Driver/Public/Mapping/MappingFailedException.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Public/Mapping/MappingFailedException.cs
@@ -13,18 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-namespace Neo4j.Driver.Preview.Mapping;
+using System;
 
-/// <summary>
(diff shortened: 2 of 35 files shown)


Neo4j.Driver/Neo4j.Driver/Public/Mapping/MappingFailedException.cs
```diff
@@ -15,7 +15,7 @@
 
 using System;
 
-namespace Neo4j.Driver.Preview.Mapping;
+namespace Neo4j.Driver.Mapping;
 
 public class MappingFailedException : Neo4jException
```
thelonelyvulpes, line 20: we are missing docs on this public class

Neo4j.Driver/Neo4j.Driver/Public/Mapping/IMappingBuilder.cs
```diff
@@ -16,7 +16,7 @@
 using System;
 using System.Linq.Expressions;
 
-namespace Neo4j.Driver.Preview.Mapping;
```
thelonelyvulpes, line 19: Neo4j.Driver.Mapping.IMappingBuilder<TObject>.Map<TProperty> is missing documentation for `optional` parameter

thelonelyvulpes (approved):
We just need to add a few docs for release

AndyHeap-NeoTech (approved):
The find replace appears to have functioned as advertised.

darcome:
Hello, where can I find docs about mapping? Thanks in advance!

RichardIrons-neo4j:
@darcome [Here is a link](https://neo4j.com/docs/api/dotnet-driver/current/api/Neo4j.Driver.Mapping.html) to the API documentation for mapping, which is correct and up to date. We do not have full documentation for the mapping functionality yet, however you can look through some of the pull requests which contain some information. Note that this documentation is now out of date in some regards.

First addition of mapping: https://github.com/neo4j/neo4j-dotnet-driver/pull/729

Addition of constructor mapping: https://github.com/neo4j/neo4j-dotnet-driver/pull/751

And then these changes made for 5.20 are that the mapping is now case sensitive in matching C# symbols to database fields and properties, and that mapping will fail if a property is not found in the record. I hope this information gets you started; feel free to start a discussion if you need help with it.

darcome:
Thank you!

PR #788, opened 2024-03-18
add handling for tcp handshake

thelonelyvulpes:
This change reclassifies the error thrown when establishing an SSL Connection.


RichardIrons-neo4j (approved):
LGTM

PR #789, opened 2024-03-20
Injectable tls negotiation

RichardIrons-neo4j:
This PR adds injectable TLS negotiation. Use one of the overloads of `WithTlsNegotiator` on the driver config builder to specify your own custom TLS negotiation. This option may compromise your application’s security if used improperly. Its usage is strongly discouraged and comes without any guarantees.


AndyHeap-NeoTech (approved):
Looking good to me.

PR #791, opened 2024-03-27
[4.4] ZoneDateTime comparison

thelonelyvulpes:
when we compare zoned date times we should use the instant values and the zone to decide if they are equal as when ambiguous times are specified they can change when sent to the server


RichardIrons-neo4j (approved):
Sorry this took a minute, lgtm.

PR #795, opened 2024-04-11
Check IP6 resolve before running test

RichardIrons-neo4j:
[::1] was hardcoded as the IP6 address to try to connect to, which will fail if as the DBMS isn't running on the same host. This was changed to use Neo4jDefaultInstallation.BoltUri instead. In addition the test is skipped if Neo4jDefaultInstallation.BoltHost cannot be resolved to an IPv6 address.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/IntegrationTestAttribute.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/IntegrationTestAttribute.cs
@@ -14,7 +14,11 @@
 // limitations under the License.
 
 using System;
+using System.Linq;
+using System.Net;
+using System.Net.Sockets;
 using System.Text;
+using Neo4j.Driver.IntegrationTests.Internals;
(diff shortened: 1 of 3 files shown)


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/IntegrationTestAttribute.cs
```diff
@@ -295,6 +299,32 @@ public sealed class RequireServerWithIPv6FactAttribute : RequireServerFactAttrib
             {
                 Skip = "IPv6 is disabled";
             }
+
+            if (string.IsNullOrEmpty(Skip))
```
thelonelyvulpes, line 303: This seems a little dangerous to me.
If we would fail the test because we mistakenly made it so we can't resolve an address it would be skipped and the suite would pass.

thelonelyvulpes (approved):
I think we need to get this in for getting out a release but further investigation is required.

PR #796, opened 2024-04-30
Make property mapping fail when record is missing field, unless the property is marked with an attribute

RichardIrons-neo4j:
This changes the behaviour of property mapping so that if a property is in the class but it cannot be populated, an exception is thrown, instead of it silently doing nothing. Two new attributes have been added:

* `[MappingOptional]` (properties only)
  This marks the property as optional. If a value is not found in the record, the mapper does nothing (the old default behaviour).
* `[MappingDefaultValue(someValue)]` (properties and constructor parameters)
  This provides a default value which will be put into the property or constructor if a suitable value is not found in the record.

Tests were added for the new features, as well as the specific issues raised that led to this change.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Preview/Mapping/MappingFailedException.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Preview/Mapping/MappingFailedException.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) "Neo4j"
+// Neo4j Sweden AB [https://neo4j.com]
+// 
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Neo4j.Driver.Preview.Mapping;
+
+public class MappingFailedException : Exception
+{
+    // the standard constructors for an exception
+    public MappingFailedException(string message) : base(message)
+    {
+    }
+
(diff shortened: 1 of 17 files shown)


Neo4j.Driver/Neo4j.Driver/Preview/Mapping/MappingFailedException.cs
```diff
@@ -0,0 +1,30 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Neo4j.Driver.Preview.Mapping;
+
+public class MappingFailedException : Exception
```
thelonelyvulpes: Should this inherit from Neo4jException?

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/MappingOptionalAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Neo4j.Driver.Preview.Mapping;
+
+/// <summary>
+/// If a property or is decorated with this attribute, it will be considered optional. The
+/// mapper will not throw an exception if it cannot find the named value in the record. This
+/// attribute will have no effect when using custom-defined mappers.
+/// </summary>
+[AttributeUsage(AttributeTargets.Property)]
+public class MappingOptionalAttribute : Attribute
```
thelonelyvulpes, line 26: Do we want to seal these attributes to avoid users inheriting from them.

Neo4j.Driver/Neo4j.Driver/Preview/Mapping/MappingFailedException.cs
AndyHeap-NeoTech: I think it would make sense that it does. I don't think there are any other exceptions that don't.

RichardIrons-neo4j: Done.

AndyHeap-NeoTech (approved):
I like what I see.

PR #810, opened 2024-08-19
Delegate mapping and blueprint mapping added to object mapping feature

RichardIrons-neo4j:
This change adds two new features to the object mapping functionality, aimed at making it easier to use in more scenarios, particularly when the developer wishes to map query results to anonymous types.

### Delegate mapping

This feature allows you to pass a delegate to the `AsObject` method to control over what happens during mapping.

As an example, let's say we have a record that looks like this:

| Field | Value |
| ---      | ---       |
| Name  | Alice |
| Age | 35 |

We can easily use delegate mapping to turn the results into a string:

```c#
var sentence = record.AsObject((string Name, int Age) => $"{Name} is {Age} years old.");
```

or more likely, an anonymously-typed object:

```c#
var obj = record.AsObject((string Name, int Age) => new { Name, Age });
```

The way this works is that the mapper looks at the names and types of the parameters that your delegate accepts (in this case `string Name, int Age`) and finds fields in the record with the same name. It then coverts the values to the correct types and calls your delegate. This of course means that you must specify the parameter types in your delegate. If any of this fails, an exception is thrown.

You can pass delegates that accept up to ten parameters.

### Blueprint mapping

This feature lets you pass an instance, rather than a type, as a "blueprint" for what object the mapper should create. For example:

```c#
var object = record.AsObjectFromBlueprint(new { Name = "", Age = 0 });
```

Again, the names and types are used to decide how to do the mapping. The values passed are never examined; only the type information is used, so whatever values were in the `Name` and `Age` properties in the previous example will be ignored.

### Notes

In both of these cases, the column names are case sensitive. If you wish to change the case of columns then you will need to use delegate mapping, for example:

```c#
var object = record.AsObject((string country_name, int population) => new { CountryName = country_name, Population = population });
```

Both pieces of functionality allow you to specify the types of your parameters/properties as anything you like, and the mapper will try to map the field to that type before passing it on. This will be done according to the global mapping configuration, so if you have custom mappers set for those type, they will be used.


Neo4j.Driver/Neo4j.Driver/Preview/Mapping/DelegateAsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,276 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Neo4j.Driver.Preview.Mapping;
+
+public static class DelegateAsyncEnumerableExtensions
```
thelonelyvulpes, line 28: Requires documentation.

RichardIrons-neo4j: Done

thelonelyvulpes (approved):

AndyHeap-NeoTech (approved):
I like the addition. I agree with the offline conversation that using a seperate method for the FromBluePrint is the way to do it rather than an overload.

michael-simons:
That's really nice.

PR #821, opened 2024-10-09
Fix default mapper bug

RichardIrons-neo4j:
This fixes a design flaw that meant when a user used the `.UseDefaultMapping()` method when building a mapper, and then specifically mapped a property using `.Map()`, the default mapper would run as normal - so if the default mapper does not know how to map the property that has been overridden, it will throw an exception before the custom mapping gets to run.

This was solved by keeping track of which properties have been explicitly mapped during building of the mapper, and then passing this list to the creation of the default mapper so that it can exclude these properties from its process.

A couple of tests have been added to verify this. 

Also there was an untested method in the mapping of an asynchronous list of records to blueprinted objects, so I put a couple in for that too.


AndyHeap-NeoTech (approved):
Approved

PR #823, opened 2024-11-06
Bump to 5.26.1

RichardIrons-neo4j:


Neo4j.Driver/common.props
```diff
@@ -1,10 +1,10 @@
 <Project>
     <PropertyGroup>
         <Copyright>Copyright 2002-2023</Copyright>
```
robsdedude, line 3: Unrelated: should this also be bumped at some point?
If so, maybe in a separate PR.

Neo4j.Driver/common.props
robsdedude, line 1: Why are we jumping from 5.25.0 to 5.26.1? Where did 5.26.0 go?

Neo4j.Driver/common.props
RichardIrons-neo4j: Yeah probably.

Neo4j.Driver/common.props
RichardIrons-neo4j: I forgot to merge the GQL errors PR before releasing 5.26.0, as well as doing this version bump. 😳

robsdedude (approved):
:bubbles:

PR #830, opened 2024-12-17
Clarify that Driver creation methods lacking auth token expect disabled auth on the server side

stefano-ottolenghi:


RichardIrons-neo4j (approved):
lgtm

Issue #831, opened 2024-12-24
`.WithNotifications()` signature broke after GQL


stefano-ottolenghi:
There seems to be a bug in notifications filtering. The GQL overloading of `.WithNotifications()` broke its intended signature. As API docs read, I should be able to call `.WithNotifications(Severity.Warning, null)`,  but that fails with

```
/home/stefano/dotnet-neo4j/Program.cs(7,105): error CS0121: The call is ambiguous between the following methods or properties: 'ConfigBuilder.WithNotifications(Severity?, Category[])' and 'ConfigBuilder.WithNotifications(Severity?, Classification[])' [/home/stefano/dotnet-neo4j/dotnet-neo4j.csproj]
```

stefano-ottolenghi:
Fixed in #834.

PR #834, opened 2025-01-09
Fix for Stefan's category/classification issue in WithNotifications

RichardIrons-neo4j:
This is a fix for [this issue](https://github.com/neo4j/neo4j-dotnet-driver/issues/831) Stefano raised.

At present, there are two overloads for the `WithNotifications` method on both the ConfigBuilder and SessionConfigBuilder:

```c#
// originally present
public ConfigBuilder WithNotifications(Severity? minimumSeverity, Category[] disabledCategories)

// added with the GQL Notifications work (which is in preview)
public ConfigBuilder WithNotifications(Severity? minimumSeverity, Classification[] disabledClassifications)
```

However, by adding the second overload, the following code will no longer compile:

```c#
builder.WithNotifications(Severity.Warning, null);
```
with the compiler giving the following error:
```
error CS0121: The call is ambiguous between the following methods or properties:
'ConfigBuilder.WithNotifications(Severity?, Category[])' and
'ConfigBuilder.WithNotifications(Severity?, Classification[])'
```

To solve the problem we can move or change the second overload since it is preview but code calling the first overload in whatever manner must compile with no changes.

The solution I went with was to combine both methods into a single overload with the following signature:

```c#
public ConfigBuilder WithNotifications(
        Severity? minimumSeverity,
        Category[] disabledCategories = null,
        Classification[] disabledClassifications = null)
```

This can be used as follows:

```c#
// same behaviour as before
.WithNotifications(Severity.Warning);

// same behaviour as before, although the IDE will point out that the
// second parameter is not required
.WithNotifications(Severity.Warning, null);

// same behaviour as before
.WithNotifications(Severity.Warning, [Category.Hint, Category.Unsupported]);

// also allowed, identical to previous line
.WithNotifications(
    Severity.Warning, 
    disabledCategories: [Category.Hint, Category.Unsupported]);

// to disable by classification
.WithNotifications(
    Severity.Warning, 
    disabledClassifications: [Classification.Hint, Classification.Unsupported]);

// or
.WithNotifications(
    Severity.Warning, 
    null,
    [Classification.Hint, Classification.Unsupported]);

// you can even do this
.WithNotifications(
    Severity.Warning, 
    [Category.Performance, Category.Topology],
    [Classification.Hint, Classification.Unsupported]);

```

In the last example, all four of the specified exclusions will apply.

I added a test that initially failed to compile, by passing null as the second parameter, and that test now works fine.

I also did some minor tidying re XML docs etc, and moved GQL Notifications out of preview.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConfigTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConfigTests.cs
@@ -294,10 +294,12 @@ public class ConfigTests
             var config = configBuilder.Build()
                 .NotificationsConfig.Should()
                 .BeOfType<NotificationsConfig>();
+
             config
                 .Which
                 .DisabledCategories.Should()
                 .BeEquivalentTo([outCat]);
+
             config
                 .Which
                 .MinimumSeverity.Should()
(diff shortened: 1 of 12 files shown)


AndyHeap-NeoTech (approved):
Looks to be correct to me.

stefano-ottolenghi (approved):
Wonderful :cartwheeling:

PR #836, opened 2025-01-19
Fluent assertions lock 7.0

thelonelyvulpes:
Fluent Assertions had its license change, this will lock the package to 7.0.0 to avoid ever going to 8.0.0 by accident.


AndyHeap-NeoTech (approved):
I saw on Friday that version 8 had moved to a payment model :o Thanks for the PR... although it's an interesting choice for a way to relax on a Sunday ;)

PR #837, opened 2025-02-13
Feature/new handshake

AndyHeap-NeoTech:
Implementation of the new manifest style handshake as per the ADR - 30 (https://github.com/neo-technology/drivers-adr/blob/main/adrs/030-handshake-manifest-v1.md)

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolVersion.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolVersion.cs
@@ -18,11 +18,13 @@ using System;
 namespace Neo4j.Driver.Internal.Protocol;
 
 //TODO: Consider converting to struct.
-internal sealed class BoltProtocolVersion : IEquatable<BoltProtocolVersion>
+internal sealed class BoltProtocolVersion : IEquatable<BoltProtocolVersion>, IComparable<BoltProtocolVersion>
 {
-    // The int 1213486160 is 0x‭48 54 54 50 - or HTTP in ascii codes... this determines the max major and minor versions supported.
-    public const int MaxMajorVersion = 80;
-    public const int MaxMinorVersion = 84;
+    // The int 1213486160 is 0x‭48 54 54 50 - or HTTP in ascii codes... determines the reserved version. It should never be sent by the server
+    private const int MaxMajorVersion = 255;
+    private const int MaxMinorVersion = 255;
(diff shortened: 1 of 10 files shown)


AndyHeap-NeoTech:
[[dotnet] Handshake Improvements implementation](https://trello.com/c/kp5Ug35w/542-dotnet-handshake-improvements-implementation)

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolVersion.cs
```diff
@@ -99,6 +105,11 @@ public bool Equals(BoltProtocolVersion rhs)
         return _compValue == rhs._compValue;
     }
 
+    public int CompareTo(BoltProtocolVersion other)
+    {
+        return other == null ? 1 : _compValue.CompareTo(other._compValue);
```
RichardIrons-neo4j, line 112: A comment would be nice on this line since the 1 is a magic number

Neo4j.Driver/Neo4j.Driver/Internal/Util/VarLong.cs
```diff
@@ -0,0 +1,38 @@
...
+    private const string ExceptionMessage = "VarLong Segment overflow";
+
+    public void AddSegment(long segment)
+    {
+        if (_position > 7)
+        {
+            throw new ArgumentException(ExceptionMessage);
+        }
+
+        segment &= 0x7F;
```
RichardIrons-neo4j, line 33: why is the parameter a long if we do this straight away, seems like it ought to be a byte?

RichardIrons-neo4j (approved):

Neo4j.Driver/Neo4j.Driver/Internal/Util/VarLong.cs
AndyHeap-NeoTech: Turns out it doesn't need to be. Have changed to Byte and adjusted the logic accordingly.

Neo4j.Driver/Neo4j.Driver/Internal/Protocol/BoltProtocolVersion.cs
AndyHeap-NeoTech: This method isn't actually used, looks like I left it in by accident. Removing.

RichardIrons-neo4j (approved):
All good!

PR #839, opened 2025-03-25
Custom type conversion added to object mapping

RichardIrons-neo4j:
This PR adds custom type conversion to the object mapping function of the driver.

Let's say you have two classes, `Customer` and `Order`, like so:

```c#
class User
{
    string Name { get; set; }
    Guid Id { get; set; }
}

class Order
{
    string Description { get; set; }
    Guid Id { get; set; }
}
```

And let's say that the `Id` values are stored in the database as string representations of GUIDs.

Previously, in order to map the string representations in the database to the Guid properties in the C# type, you would need to define a custom mapper for both the `User` and `Order` types. And indeed, any other types that had a Guid property that was going to need to be mapped from a string.

With this PR, you can provide a conversion to the mapping config so that if it ever needs a Guid and it has a string, it will call your provided conversion. At the time of setting up the driver, add the following code:

```c#
RecordObjectMapping.RegisterTypeConverter<string, Guid>(s => Guid.Parse(s));
```

Now, whenever it needs a Guid and it has a string, the mapper will know what to do. This will work with property mapping, constructor mapping, delegate mapping, blueprint mapping, and all other kinds of mapping functionality.

The example here is simple, but more complex scenarios could be handled with ease. For example, if the database has values stored as JSON, this feature could be used to map those values directly to typed objects.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Internal/Util/VarLongTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Internal/Util/VarLongTests.cs
@@ -15,7 +15,6 @@
 
 using System;
 using FluentAssertions;
-using Neo4j.Driver.Internal.IO;
 using Neo4j.Driver.Internal.Util;
 using Xunit;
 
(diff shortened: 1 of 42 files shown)


AndyHeap-NeoTech (approved):
Nice work. I can't see any obvious issues in the implementation. API is clean and easy to use :)

PR #840, opened 2025-03-28
Naming convention translation for object mapping

RichardIrons-neo4j:
## Naming Convention Translation in Object Mapping

Naming conventions are often different in C# to what they are in the database, even though the names themselves are
basically the same. It could be something as simple as mismatched case, for example a property on a node in the database
called `title` in the database when the equivalent C# property is called `Title`, or it could be something more complex
like the property in the database being called `number_of_units` versus a C# constructor parameter called numberOfUnits.

Because the object mapping functionality requires strict matching to avoid unexpected nodes, this can mean decorating
every property or parameter with a `[MappingSource]` attribute, for example:

```c#
public class Example
{
    [MappingSource("exampleProperty")]
    public string ExampleProperty { get; set; }
    // ...
}
```

This PR introduces a method to define a translation from C# identifiers to the identifiers in the database that will be
applied in all situations except if a `[MappingSource]` attribute is used. This is set up once and applied to all
mapping operations thereafter.

### Parsing and Formatting

The translation works by first _parsing_ some set of data (typically a list of tokens) from the C# identifier and then _formatting_ this data to form a
new identifier, which is the one looked for in the database.

In most cases, the C# identifier will conform to normal C# standards - beginning with either an upper or lower case
character, and then each new token beginning with a capital letter, and the database field names will be in camel
case - beginning with a lower case character and then each new token beginning with a capital letter. For example, the C# property `NumberOfUnits` would be represented in the database as `numberOfUnits`.

To do this simple conversion, simply call the `RecordObjectMapping.TranslateIdentifiers()` method with no parameters, which will set up the default translation.

To translate between common casing conventions, call the `RecordObjectMapping.TranslateIdentifiers` method passing either a `FieldCaseConvention`, an `IdentifierCaseConvention`, or both. 

Example configuration calls for various case conventions:

| Identifier Name   | Field Name             | Call to `RecordObjectMapping.TranslateIdentifiers`                          |
|-------------------|------------------------|-----------------------------------------------------------------------------|
| `NumberOfUnits`   | `numberOfUnits`        | `RecordObjectMapping.TranslateIdentifiers()`                                |
| `NumberOfUnits`   | `number_of_units`      | `RecordObjectMapping.TranslateIdentifiers(FieldCaseConvention.SnakeCase)`  |
| `NumberOfUnits`   | `number-of-units`      | `RecordObjectMapping.TranslateIdentifiers(FieldCaseConvention.KebabCase)`  |
| `NumberOfUnits`   | `NUMBER_OF_UNITS`      | `RecordObjectMapping.TranslateIdentifiers(FieldCaseConvention.ScreamingSnakeCase)` |
| `numberOfUnits`   | `number_of_units`      | `RecordObjectMapping.TranslateIdentifiers(IdentifierCaseConvention.CamelCase, FieldCaseConvention.SnakeCase)` |
| `NumberOfUnits`   | `number-of-units`      | `RecordObjectMapping.TranslateIdentifiers(IdentifierCaseConvention.PascalCase, FieldCaseConvention.KebabCase)` |
| `number_of_units` | `numberOfUnits`        | `RecordObjectMapping.TranslateIdentifiers(IdentifierCaseConvention.SnakeCase)` |
| `NUMBER_OF_UNITS` | `numberOfUnits`        | `RecordObjectMapping.TranslateIdentifiers(IdentifierCaseConvention.ScreamingSnakeCase)` |

An exception will be thrown if the identifier does not conform to the standard.

If you want to be in complete control of the translation, you can bypass the extraction and combination
steps completely and provide your own translator by implementing the `IConventionTranslator` interface. This 
interface has a single method, `Translate`, which takes a string and returns a string. You would then call 
`SetConventionTranslation(new MyTranslator())`. For example:

```c#
public class MyTranslator : IConventionTranslator
{
    // This example prepends "db_" to the input string
    public string Translate(string input)
    {
        return "db_" + input;
    }
}

//...

public class Example
{
    public string ExampleProperty { get; set; }
    // ...
}
    
RecordObjectMapping.SetConventionTranslation(new MyTranslator());

var result = driver
    .ExecutableQuery("MATCH (n:Node) RETURN n.exampleProperty AS db_ExampleProperty")
    .ExecuteAsync()
    .AsObjects<Example>();
```
The mapper will now look for `db_ExampleProperty` in the database when trying to find a value for `ExampleProperty`.

If any properties or parameters are marked with a `[MappingSource]` attribute, the translation will not be applied to
those properties or parameters. This allows you to override the translation for specific properties or parameters.


MaxAake:
1 small comment on the implementation on the token extractors. I feel like it'd make making a Combiner easier if the Extractors all produced identical tokens (all lower-case, for example). As is, most combiners would need to start with a conversion to lower case.

I am wondering if we should have a talk to standardize the names for these functions on all drivers, might be useful to ensure that names do not differ too much?

Other than those two small notes I like the implementation! It's very similar to the approach I took in the JS driver.

RichardIrons-neo4j:
> 1 small comment on the implementation on the token extractors. I feel like it'd make making a Combiner easier if the Extractors all produced identical tokens (all lower-case, for example). As is, most combiners would need to start with a conversion to lower case.

I did think about this, but in the end, I decided against it, purely because I didn't want to have a contract in place that wouldn't be enforced by the compiler. As it is, the rule is "you cannot assume anything about the case of the tokens". If we changed it so that the rule was "the tokens will always be lower case", then we might end up in a situation where they write their own extractor which returns mixed-case tokens, and then they use a combiner which is assuming the tokens will be lower case, leading to undefined results. 

> I am wondering if we should have a talk to standardize the names for these functions on all drivers, might be useful to ensure that names do not differ too much?

I like this idea.

AndyHeap-NeoTech (approved):
Lovely stuff. Pre-approved pending any naming changes to the methods once discussed with the team.

MaxAake:
Context for the discussion next week:

In the current JavaScript draft, there are currently the functions: `neo4j.mapping.translatePropertyNames()` and `neo4j.mapping.defaultNameTranslation()`, both with names I am rather unhappy with.

`neo4j.mapping.translatePropertyNames()` takes a function to be used to map all property names without a set mapping source. Therefore, this is fairly equivalent to the dotnet `RecordObjectMapping.SetConventionTranslation();`

`neo4j.mapping.defaultNameTranslation()` takes two strings, each a naming convention. supported conventions are "camelCase", "PascalCase", "snake_case", "SNAKE_CAPS" and "kebab-case". The function returns a translation function that can be passed into`neo4j.mapping.translatePropertyNames()`. 

To set a translation for PascalCase in code and camelCase in the database, the user would therefore write
```javascript
neo4j.mapping.translatePropertyNames(neo4j.mapping.defaultNameTranslation("camelCase", "PascalCase"))
```

Neo4j.Driver/Neo4j.Driver.Tests/Mapping/ConventionTranslation/FormatterTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using FluentAssertions;
+using Neo4j.Driver.Mapping.ConventionTranslation;
+using Xunit;
+
+namespace Neo4j.Driver.Tests.Mapping.ConventionTranslation;
+
+public class FormatterTests
+{
+    [Fact]
+    public void SnakeCaseCombiner_ShouldFormat()
```
MaxAake:
```suggestion
    public void SnakeCaseFormatter_ShouldFormat()
```

Nitpick. The Formatter and Identifier tests still use "extractor" and "combiner". Not a blocker so I will approve, up to you if you feel the need to amend.

MaxAake (approved):
Changes to be in line with the discussion on naming look good to me! One comment on the naming of tests.

Issue #844, opened 2025-04-24
Bug in new version (HOME Db CACHE)


killJoy-03:
The LinkedList node does not belong to current LinkedList.

at Neo4j.Driver.Internal.AsyncTransaction.TransactionConnection.OnErrorAsync(Exception error)
   at Neo4j.Driver.Internal.Connector.DelegatedConnection.SyncAsync()
   at Neo4j.Driver.Internal.Protocol.BoltProtocolV3.BeginTransactionAsync(IConnection connection, BeginTransactionParams beginParams, HomeDbCacheKey cacheKey, IHomeDbCache homeDbCache, SessionConfig sessionConfig)
   at Neo4j.Driver.Internal.Protocol.BoltProtocol.BeginTransactionAsync(IConnection connection, BeginTransactionParams beginParams, HomeDbCacheKey cacheKey, IHomeDbCache homeDbCache, SessionConfig sessionConfig)
   at Neo4j.Driver.Internal.AsyncSession.BeginTransactionWithoutLoggingAsync(AccessMode mode, TransactionConfig config, Boolean disposeUnconsumedSessionResult, TransactionInfo transactionInfo)
   at Neo4j.Driver.Internal.AsyncSession.<>c__DisplayClass54_0`1.<<RunTransactionAsync>b__1>d.MoveNext()
--- End of stack trace from previous location ---
   at Neo4j.Driver.Internal.AsyncRetryLogic.RetryAsync[T](Func`1 runTxAsyncFunc)
   at Neo4j.Driver.Internal.Logging.DriverLoggerUtil.TryExecuteAsync[T](ILogger logger, Func`1 func, String message)

killJoy-03:
v5.28.0

RichardIrons-neo4j:
Hi @killJoy-03 - thanks for reporting this bug so promptly, and my apologies that it wasn't caught in testing. I've [just put in a PR](https://github.com/neo4j/neo4j-dotnet-driver/pull/845) with a fix for this, and a patch release should be out on Monday.

RichardIrons-neo4j:
Fixed in latest patch release.

PR #845, opened 2025-04-25
Fix for https://github.com/neo4j/neo4j-dotnet-driver/issues/844

RichardIrons-neo4j:
Added tests looking for the bug. The failing test was the multithreaded one. On checking, the home DB cache was not thread-safe (my bad). Fixed it and the test passes. This will go straight into a patch release.


AndyHeap-NeoTech (approved):
LGTM

PR #851, opened 2025-07-01
Fix parameter bug and enhance documentation for WithTlsNegotiator method

RichardIrons-neo4j:
Added specific tests for the case that caused the bug. Fixed using a struct wrapper for the dictionary that won't cause any allocations. Also a little docs update for specifying your own ITlsNegotiator.


MaxAake (approved):
✅

PR #856, opened 2025-07-17
ResultCursorExtensions cancellation support

Sossenbinder:
I merely added cancellation propagation support for all methods in ResultCursorExtensions, according to what is already applied to the `ToListAsync()` extension.

We're currently in the progress of adding cancellation propagation through Neo4J in our own repo using the driver, but we've noticed that only a few of the result helpers expose a CancellationToken as of now, despite all of them using the very same Result async enumerator with complete cancellation support.


RichardIrons-neo4j:
This work will be in 6.

PR #859, opened 2025-08-08
Vector support

RichardIrons-neo4j:
Introduce Bolt 6.0 support, which includes vector types.


Neo4j.Driver/Neo4j.Driver.Tests.TestBackend/Protocol/Session/SessionRun.cs
```diff
@@ -38,15 +39,26 @@ public override async Task Process()
...
+        }
 
         ResultId = result.uniqueId;
     }
 
     public override string Respond()
     {
-        return ((Result.Result)ObjManager.GetObject(ResultId)).Respond();
+        var protocolObject = (Result.Result)ObjManager.GetObject(ResultId);
+        var type = protocolObject.GetType();
```
AndyHeap-NeoTech: Type is unused. Looks like code was changed for some debugging.

Neo4j.Driver/Neo4j.Driver.Tests/Mapping/MappingProviderTests.cs
```diff
@@ -102,17 +102,6 @@ public void ShouldNotFailWhenUsingDefaultMapperButMappingSomePropertiesExplicitl
         obj.Guid.Should().Be(guid);
     }
 
-    [Fact]
```
AndyHeap-NeoTech, line 105: Just wondering what the reason for removing this test is?

Neo4j.Driver/Neo4j.Driver/Public/Types/IVector.cs
```diff
@@ -0,0 +1,51 @@
...
+/// Supported element types are: <see cref="float"/>, <see cref="double"/>, <see cref="sbyte"/>, <see cref="short"/>,
+/// <see cref="int"/>, and <see cref="long"/>.
+/// </remarks>
+public interface IVector
+{
+    /// Returns the elements of the vector as an array of objects, regardless of their underlying type.
+    IEnumerable<object> UntypedValues { get; }
+
+    /// Gets the original byte stream from which the vector was deserialized, if applicable.
+    byte[] OriginalByteStream { get; }
```
AndyHeap-NeoTech, line 34: Is it necessary to store the original byte stream? It looks to be unused, I can't see an instance of the accessor being called, and the byte stream can be reconstructed from teh UntypedValues if needed. I'm thinking about the impact on memory footprint here.

Neo4j.Driver/Neo4j.Driver/Public/Types/Vector.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// <summary>
+    /// Gets the elements of the vector as an array of objects, regardless of their underlying type.
+    /// </summary>
+    public abstract IEnumerable<object> UntypedValues { get; }
+
+    /// <summary>
+    /// Gets the original byte stream from which the vector was deserialized, if applicable.
+    /// </summary>
+    public abstract byte[] OriginalByteStream { get; }
```
AndyHeap-NeoTech, line 65: Perhaps we should convert from the UntypedValues here rather than store the bytestream

Neo4j.Driver/Neo4j.Driver.Tests/Mapping/MappingProviderTests.cs
RichardIrons-neo4j: It's because as a 6.0 feature GUIDs are automatically translated from a string if possible. Not vector related I know - could move that change to a separate PR if you think it's needed. I forgot I did this tbh.

AndyHeap-NeoTech: Nah no need to move it.

Neo4j.Driver/Neo4j.Driver/Public/Types/IVector.cs
RichardIrons-neo4j: I've made a change here so that the original byte stream that's stored isn't a copy of the array, it's just the original array. The switching of endianness is now done in a temporary copy of the array on the stack.

Neo4j.Driver/Neo4j.Driver/Public/Types/Vector.cs
RichardIrons-neo4j: I wouldn't want to reverse-engineer the byte stream from the untyped values, I don't think.

AndyHeap-NeoTech (approved):
I think that looks good to go.

RichardIrons-neo4j:
Closed by accident.

PR #861, opened 2025-08-15
Improved IsRetriable method reference description

AndyHeap-NeoTech:
Improved IsRetriable method reference description to highlight that the flag does not apply safely to Session.Run because of things like periodic commit and call in tx.

Tidied up some compiler warning from incomplete or erroneous XML comments.


StephenCathcart (approved):
👍 LGTM

PR #862, opened 2025-08-29
Added missing methods to two public interfaces that should have been there

AndyHeap-NeoTech:
1. BeginTransaction - It was not possible to do this whilst specifying an AccessMode to override the sessions configured default. Two signatures added in the IAsyncSession interface and implementations in the concrete AsyncSession to enable this.

2. GetExecutableQueryBookmarkManager - added to IDriver interface and an implementation in the Driver class. This allows access to the bookmark manager used in ExecutableQueries so it can be passed to sessions and allow causal consistency between the two ways of running queries.


RichardIrons-neo4j (approved):
lgtm

PR #863, opened 2025-08-29
Fixing two interfaces missing methods and updated some reference documentation.

AndyHeap-NeoTech:
Fixing two interfaces missing methods and updated some reference documentation.

BeginTransaction - It was not possible to do this whilst specifying an AccessMode to override the sessions configured default. Two signatures added in the IAsyncSession interface and implementations in the concrete AsyncSession to enable this.

GetExecutableQueryBookmarkManager - added to IDriver interface and an implementation in the Driver class. This allows access to the bookmark manager used in ExecutableQueries so it can be passed to sessions and allow causal consistency between the two ways of running queries.


RichardIrons-neo4j (approved):
also fine

PR #866, opened 2025-09-19
Fix for new handshake and unit test.

AndyHeap-NeoTech:
Fix for a handshake unit test that was failing once the version was increased to 6.0. This was not actually a failing test but an error in the test itself. On the second protocol entry in sent by the 'server' it supplied a version instead of a range. This meant that under certain circumstances the test would pass, which just so happened to be the case with the current versions when the handshake was added.


RichardIrons-neo4j (approved):
All good.

PR #871, opened 2025-10-02
Fix for incorrect reporting of a type mismatch when object mapping fails and a Vector type is involved.

AndyHeap-NeoTech:
Before the fix the error reported when trying to map from, for example, a Vector<float> to a Vector<double> was: 

`System.InvalidOperationException: The expected value 'Neo4j.Driver.Vector'1[System.Single]' is different from the actual value 'System.Collections.Generic.List'1[System.Single]' `

This was occurring because the ValueExtensions As method did not contain a clause for IVector, this caused the logic to fall through to the IEnumerable clause (which a vector implements) which converted the vector into a list, and then failed the type conversion on that list.

Now that a clause for IVector is added you get the correct error. e.g. For the Vector<float> -> Vector<double> example: 

`System.InvalidCastException: Cannot cast differing vector types. Attempting from Neo4j.Driver.Vector'1[System.Double]' to 'Neo4j.Driver.Vector'1[System.Single]`

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ValueExtensionsTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ValueExtensionsTests.cs
@@ -236,6 +236,30 @@ public class ValueExtensionsTests
             var ex = Record.Exception(() => value.As<int>());
             ex.Should().BeOfType<InvalidCastException>();
         }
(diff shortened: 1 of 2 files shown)


RichardIrons-neo4j (approved):
👌

PR #872, opened 2025-10-03
Vector bugfix

RichardIrons-neo4j:
Serialisation of parameters was failing on Vectors.

Refactored the parameter transformation to allow better testing, added failing test, fixed bug. The Transform method now returns the value immedi


Neo4j.Driver/Neo4j.Driver/Internal/Util/ObjectToDictionary/ParameterValueTransformer.cs
```diff
@@ -0,0 +1,180 @@
...
+namespace Neo4j.Driver.Internal.Util;
+
+internal class ParameterValueTransformer : IParameterValueTransformer
+{
+    private static readonly TypeInfo NeoValueTypeInfo = typeof(IValue).GetTypeInfo();
+
+    public object Transform(object value)
+    {
+        var valueType = value?.GetType();
+        if (valueType == null || !NeedsConversion(valueType))
```
AndyHeap-NeoTech, line 32: Am I correct in thinking that the core of the fix for the bug is the addition of the call to NeedsConversion here? Hard to see relative changes in a refactor :(

AndyHeap-NeoTech (commented):
One question added.

Neo4j.Driver/Neo4j.Driver/Internal/Util/ObjectToDictionary/ParameterValueTransformer.cs
RichardIrons-neo4j: Yeah that's right, it was the only thing. But it was hard to make a failing test for it (to reproduce the bug).

AndyHeap-NeoTech (approved):
All good

PR #874, opened 2025-10-23
Fix connection acquisition timeout

RichardIrons-neo4j:
Also includes the removal of deprecated severity property.

Closes DRIVERS-62


Neo4j.Driver/Neo4j.Driver/Internal/Routing/LoadBalancer.cs
```diff
@@ -97,6 +97,18 @@ public async Task<IConnection> AcquireAsync(
         SessionConfig sessionConfig,
         Bookmarks bookmarks,
         bool forceAuth)
+    {
+        return await AcquireConnectionInternalAsync(mode, database, sessionConfig, bookmarks, forceAuth)
```
AndyHeap-NeoTech, line 101: Is it the case that because AcquireConnectionInternalAsync does not observe a cancellationToken that the underlying operations might continue running after the await timesout? The timeout is effectively only impacting the awaiting task? The ConnectionAcquisition timeout is still being used on several steps lower down the stack (), so I'm wondering if the CancelationToken should be passed down the layers that way enforcing the timeout across all operations.

PR #876, opened 2025-11-03
Removal of IPv6 config option

AndyHeap-NeoTech:
Closes DRIVERS-85

Removal of IPV6 config option. The driver will support IPv6 and IPv4 by default going forward.
The driver always runs it's sockets in dual mode now.
Note: Several tests were removed that focused on address ordering during resolution. Preferring either IPv4 or IPv6 addresses depending on config. This ordering is not changed now.


RichardIrons-neo4j (approved):
Looks good.

PR #881, opened 2025-11-10
Renamed ILogger to INeo4jLogger in preparation for integration with standard logging framework

RichardIrons-neo4j:
Closes DRIVERS-93 This change renames the ILogger interface to INeo4jLogger so that when integration with Microsoft's standard logging framework is implemented, there will not be any ambiguity. I also renamed all the classes and variables so that they'll be clear too.

There is no change of functionality in this change, just the rename.


AndyHeap-NeoTech (approved):
LGTM.

PR #885, opened 2025-11-19
Unify duration string representation and add tests

RichardIrons-neo4j:
Looked at the full ISO 8601 duration string spec and implemented that, plus the simplification, and added tests.

Closes DRIVERS-201


AndyHeap-NeoTech (approved):
Looks complete to me.

PR #893, opened 2025-12-12
Move the simple driver into a tests namespace and remove its ability to be built into a nuget package

RichardIrons-neo4j:
There will be a change to the build pipeline coming shortly.

Closes DRIVERS-88.


AndyHeap-NeoTech (approved):
Looks good to go.

PR #896, opened 2025-12-17
Remove conditional compiles

RichardIrons-neo4j:
Remove conditional compiles on the basis that .NET 8 is the latest supported version.

Closes DRIVERS-238.


AndyHeap-NeoTech (approved):
Looks good.

PR #897, opened 2025-12-18
Deprecate notifications in favor of GQL statuses 

RichardIrons-neo4j:
* Marked the `INotification` interface and the `Notifications` property in both `IResultSummary` and `ResultSummary` as obsolete.

* Deprecated the `Code` property in `FailureMessage`

* Updated the string representation of `FailureMessage` to show `gqlStatus` instead of the legacy `code` field.

* Amended test for said ToString method.

Closes DRIVERS-86.


AndyHeap-NeoTech (approved):
Looks correct. Was their a requirement to move GQLStatusObject out of preview or was that already done?

PR #898, opened 2026-01-06
Fix all warnings in the driver project

RichardIrons-neo4j:
Changes for deprecation warnings (both ours and Microsoft's) and XML documentation. Also made sure the unit test project was targeting .NET 8, 9 and 10.


AndyHeap-NeoTech (approved):
LGTM

PR #899, opened 2026-01-13
Fix preview references in xml documentation

RichardIrons-neo4j:
Some of the XML documentation around GQL statuses and notifications still said it was in preview; this has been rectified.


AndyHeap-NeoTech (approved):
Looking good.

PR #903, opened 2026-03-10
ProfiledPlan: expose existence for all metrics

robsdedude:
Deprecate `IResultSummary.Profile()` returning an `IProfiledPlan` in favor of `ResultSummary.QueryProfile` returning an `IProfile`.

The main difference between the profile interfaces is that the new one exposes whether stats (such as `DbHits`, `Time`, etc.) have been recorded by returning nullable types. Further, `IProfiledPlan.Records` corresponds to the new `IProfile.Rows` to better align with Neo4j's embedded API, Bolt, and the raw arguments attached to the profile steps.

According to a chat with the Cypher runtime team, the absence of the value is not equivalent to it being zero. Therefore it makes sense to express this information in the API. The information is encoded in the bolt protocol through absence or existence of the respective fields.

Closes: DRIVERS-288

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/705


robsdedude:
Please don't merge yet. I'd like to wait for this work to be ready for merging across all drivers first.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Direct/ResultIT.cs
```diff
@@ -145,11 +146,83 @@ public async Task GetProfile()
...
+
+            summary.HasPlan.Should().BeTrue();
+            summary.HasProfile.Should().BeTrue();
+
+            summary.Plan.Should().Be(summary.QueryProfile);
+
+            var profile = summary.QueryProfile;
+            profile.DbHits.Should().Be(0L);
+            profile.Rows.Should().Be(1L);
+            profile.Rows.Should().Be(1L);
```
RichardIrons-neo4j: duplicate line - meant to be something else?

Neo4j.Driver/Neo4j.Driver/Internal/Result/Summary/ProfiledPlan.cs
```diff
@@ -57,12 +64,15 @@ public ProfiledPlan(
...
 
+    public bool HasDbHits { get; }
     public long DbHits { get; }
 
+    public bool HasRecords { get; }
     public long Records { get; }
     public long PageCacheHits { get; }
     public long PageCacheMisses { get; }
     public double PageCacheHitRatio { get; }
+    public bool HasTime { get; }
```
RichardIrons-neo4j: HasDbHits, HasRecords and HasTime are in here but not in the public interface. Remove?

Neo4j.Driver/Neo4j.Driver/Public/Summary/IProfile.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System.Collections.Generic;
+
+namespace Neo4j.Driver;
+
+/// <summary>
+/// This is the same as a regular <see cref="IPlan"/> - except this plan has been executed, meaning it also
+/// contains detailed information about how much work each step of the plan incurred on the database.
+/// </summary>
+public interface IProfile : IPlan
```
RichardIrons-neo4j: Could we ha ve a less generic name? IQueryProfile or IProfiledQueryPlan or something?

Neo4j.Driver/Neo4j.Driver/Internal/Result/Summary/SummaryBuilder.cs
```diff
@@ -72,7 +76,10 @@ public ResultSummary(SummaryBuilder builder, CursorMetadata cursorMetadata)
         public bool HasPlan => Plan != null;
```
RichardIrons-neo4j, line 76: Not sure but I _think_: when the deprecated `IProfiledPlan Profile` is eventually removed, HasProfile will break and will need to check `QueryProfile != null`

RichardIrons-neo4j (commented):
Just a few notes/queries

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Direct/ResultIT.cs
robsdedude: Good catch. All other fields already have assertions. So this is just a useless duplicate :wastebasket:

Neo4j.Driver/Neo4j.Driver/Internal/Result/Summary/ProfiledPlan.cs
robsdedude: Yeah totally. It's a remnant of the previous design 🧹

Neo4j.Driver/Neo4j.Driver/Public/Summary/IProfile.cs
robsdedude: I will leave this one open until the other driver PRs have been reviewed. After that I suggest we'll have a discussion with the whole team. The name should be the same (or as similar as possible) across drivers with all team members being ok with the chosen name.

Neo4j.Driver/Neo4j.Driver/Internal/Result/Summary/SummaryBuilder.cs
robsdedude: Yeah, that's right. With the current code base `QueryProfile == null` is true if and only if `Profile == null` is true. So it doesn't really matter which one to choose for the implementation of `HasProfile`. If you prefer, I can switch it over to using `QueryProfile` for the null check right now. Just let me know.

RichardIrons-neo4j: Wait - it'll be a compiler error won't it. Yeah leave it.

RichardIrons-neo4j (approved):
All good, just waiting on other drivers for minor unification stuff.

PR #909, opened 2026-04-11
Bolt strangler fig stage 1

RichardIrons-neo4j:
Adds an INeo4jServer internal interface as a seam between the driver and the existing Bolt stack, with a thin BoltServer wrapper that just delegates to IConnectionProvider. Driver now talks to INeo4jServer only — no direct dependency on the connection infrastructure. 

The intent is that when the new Bolt library is ready, and when QueryAPI is ready, it slots in here behind an adapter without touching anything above it.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/AsyncSessionTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/AsyncSessionTests.cs
@@ -410,7 +410,7 @@ public class AsyncSessionTests
         private static Driver.Internal.Driver NewDriver()
         {
             var context = TestDriverContext.MockContext;
-            var server = new BoltServer(
+            var server = new BoltProtocolAdapter(
                 new TestConnectionProvider(Mock.Of<IConnection>()),
                 context);
 
--- a/Neo4j.Driver/Neo4j.Driver.Tests/AsyncSessionTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/AsyncSessionTests.cs
@@ -409,11 +409,15 @@ public class AsyncSessionTests
 
         private static Driver.Internal.Driver NewDriver()
         {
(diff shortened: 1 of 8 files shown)


AndyHeap-NeoTech:
Implementation looks good, I just want to summarise in my own words the approach here. 

You are abstracting session creation, and then the bolt implementation and queryAPI implementation will each implement their own versions of AsyncQueryRunner (I don't _think_ AsyncSession will need abstracting).

I think this works. Just a couple of things to consider, and on the surface I think this works with them as well due to separation of concerns. 
- Are we still able to swap out packstream. Yes because it sits below this abstraction
- Are we still able to swap the transport layer. Yes because it sits even further down the stack below bolt
- Are we still able to independently change bolt and queryApi. Yes, only new functionality that surfaces to the user will impact both. 

One other thought, is there anything we need to consider around ResultCursor and Records as they surface back up through the abstraction? Again, I _think_ it is ok.

RichardIrons-neo4j:
> Implementation looks good, I just want to summarise in my own words the approach here.
> 
> You are abstracting session creation, and then the bolt implementation and queryAPI implementation will each implement their own versions of AsyncQueryRunner (I don't _think_ AsyncSession will need abstracting).
> 
> I think this works. Just a couple of things to consider, and on the surface I think this works with them as well due to separation of concerns.
> 
> * Are we still able to swap out packstream. Yes because it sits below this abstraction
> * Are we still able to swap the transport layer. Yes because it sits even further down the stack below bolt
> * Are we still able to independently change bolt and queryApi. Yes, only new functionality that surfaces to the user will impact both.
> 
> One other thought, is there anything we need to consider around ResultCursor and Records as they surface back up through the abstraction? Again, I _think_ it is ok.

Yep, I agree with all that.

AndyHeap-NeoTech (approved):

PR #910, opened 2026-04-12
Idempotent retries

RichardIrons-neo4j:
CLOSES DRIVERS-311

Idempotent retry for auto-commit transactions (session.Run). When the server rejects a RUN with a FAILURE whose diagnostic record contains _idempotent: true, the driver closes the current connection, acquires a fresh one, and retries the query exactly once.

The retry is transparent to the caller — errors from both the original and retry attempts surface lazily during result iteration, consistent with existing behaviour.

A new GetRunCompletionErrorAsync method on the result cursor peeks at the response pipeline to inspect the RUN outcome without consuming the error throw - lazy error surfacing is unaffected. The feature can be opted out via DisableAutoCommitRetries at the driver or session level.

**Implementation note:** Fixed the pipeline ordering case discussed in standup — `_pipelineFailedBeforeRunResponse` is snapshotted at catch time before the connection drain can call `RunCompleted` and confuse the RUN-received check.


Neo4j.Driver/Neo4j.Driver/Internal/AsyncSession.cs
```diff
@@ -202,27 +206,53 @@ public Task<IResultCursor> RunAsync(
...
+                var cursor = await _connection
+                    .RunInAutoCommitTransactionAsync(autoCommitParams, _notificationsConfig, _driverContext?.HomeDbCache)
                     .ConfigureAwait(false);
+
+                if (disableRetries)
+                {
+                    return cursor;
+                }
+
+                if (cursor is IInternalResultCursor irc)
```
AndyHeap-NeoTech, line 233: Logically this is sound, but I'm thinking it would be more elegant and future proofed if it was done in a loop rather than repeating a lot of the code. We may want to go to more than a single retry in the future, or make changes along those lines.

AndyHeap-NeoTech (approved):
Looks good. Just one suggestion.

Neo4j.Driver/Neo4j.Driver/Internal/Extensions/ErrorExtensions.cs
```diff
@@ -35,6 +35,17 @@ public static bool CanBeRetried(this Exception error)
         return error is Neo4jException { IsRetriable: true };
     }
 
+    /// <summary>
+    /// Returns true if the server marked the failure as idempotent — i.e. guaranteed that no state
+    /// change occurred, so the request can be safely retried (e.g. admission control rejection).
+    /// </summary>
+    public static bool IsIdempotentFailure(this Exception error)
```
AndyHeap-NeoTech, line 42: I'm unsure if this should be public. We need to be consistent across drivers so need to check on that. Also it can be made public later if needed, but it's a breaking change if we want to make it private at a later date.

RichardIrons-neo4j: This is in an internal class so it won't be public.

Neo4j.Driver/Neo4j.Driver/Internal/AsyncSession.cs
RichardIrons-neo4j: Created a ticket to do this: https://linear.app/neo4j/issue/DRIVERS-341/refactor-idempotent-retry-logic-to-use-a-loop

PR #911, opened 2026-04-13
SPIKE: DRIVERS-195 improve routing error with per-router causes

RichardIrons-neo4j:
CLOSES DRIVERS-195

Surfaces per-router connection exceptions as an `AggregateException` inner on the final `ServiceUnavailableException` when all routing servers fail, so the root cause (e.g. connection refused, DNS failure) is no longer silently discarded.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,10 +1,8 @@
-# Ignore everything
-/*
+:include .gitignore
 
-# except:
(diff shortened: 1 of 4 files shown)


AndyHeap-NeoTech (approved):
Looks sensible to me.

AndyHeap-NeoTech (approved):

PR #916, opened 2026-04-14
Add GQL status finder methods to Neo4jException

RichardIrons-neo4j:
CLOSES DRIVERS-81

Adds `ContainsGqlStatus(string)` and `FindByGqlStatus(string)` to `Neo4jException`. Both methods walk the GQL cause chain via `InnerException`, same as as the Java driver (neo4j/neo4j-java-driver#1671).


AndyHeap-NeoTech (approved):
Change looks good. Is it worth linking to online documentation of the GQL status codes? Risk would be link rot in the future...

AndyHeap-NeoTech (approved):

PR #917, opened 2026-04-14
docs: improve encryption and TLS configuration API docs

RichardIrons-neo4j:
CLOSES DRIVERS-26

Improves the XML doc comments on the encryption and TLS configuration surface (`EncryptionLevel`, `CertificateTrustRule`, `TrustManager`, and the `ConfigBuilder` methods that use them).

Key changes:
- Explains that `+s`/`+ssc` URI schemes, `WithEncryptionLevel`, and `WithTrustManager`/`WithCertificateTrustRule` are mutually exclusive — and that the URI scheme approach is preferred.
- Adds a security warning to `TrustManager.CreateInsecure` and `CertificateTrustRule.TrustAny`.
- Fixes stale "We recommend using WithCertificateTrustPaths or WithCertificates" remark on `WithTrustManager` (those methods don't exist).
- Rewrites the confusingly-worded "Not to be used when using a non-basic URI scheme" remarks on `WithCertificateTrustRule`.
- Improves description of each `CertificateTrustRule` value and `TrustManager` factory method with clearer guidance on when to use each.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Public/CertificateTrustRule.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Public/CertificateTrustRule.cs
@@ -15,15 +15,33 @@
 
 namespace Neo4j.Driver;
 
-/// <summary>Rules for how to validate a server certificate.</summary>
+/// <summary>
+/// Controls which server certificates the driver will accept when establishing a TLS connection.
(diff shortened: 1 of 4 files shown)


AndyHeap-NeoTech (approved):
LGTM

PR #921, opened 2026-04-15
Backport DRIVERS-26 and DRIVERS-200 to 5.x

RichardIrons-neo4j:
Backport of two doc-only fixes to the 5.x branch.

- **DRIVERS-26**: Improve encryption and TLS configuration API docs — clarifies URI scheme vs explicit API relationship, adds security warnings to insecure trust options.
- **DRIVERS-200**: Clarify disposal behaviour on `IDriver` — makes clear that sessions/transactions must be closed before disposal to avoid unspecified behaviour.


robsdedude (approved):
🍒⛏️

PR #928, opened 2026-05-11
Add native UUID / Bolt 6.1 support

RichardIrons-neo4j:
Closes DRIVERS-358. 

Adds native UUID support for Bolt 6.1. 

UUIDs are transmitted as marker byte `0xE0` + 16 big-endian bytes and surfaced in the public API as `System.Guid`. 

Implementation covers: 
 * stream-based and span-based read paths
 * query parameters 
 * `ValueExtensions.As<Guid>()`
 * object mapping


Neo4j.Driver/Neo4j.Driver.Tests/Mapping/UuidMappingTests.cs
```diff
@@ -0,0 +1,191 @@
...
+    [Fact]
+    public void Record_TypedGet_WithKnownValue_ReturnsExpectedGuid()
+    {
+        var guid = Guid.NewGuid();
+        var record = TestRecord.Create(["id"], [guid]);
+
+        record.Get<Guid>("id").Should().Be(guid);
+    }
+
+    [Fact]
```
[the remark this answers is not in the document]
RichardIrons-neo4j: Removed duplicate test.

Neo4j.Driver/Neo4j.Driver/Public/Extensions/ValueExtensions.cs
```diff
@@ -145,6 +145,16 @@ public static T As<T>(this object value)
             return Convert.ToBoolean(value).AsItIs<T>();
         }
 
+        if (targetType == typeof(Guid))
```
AndyHeap-NeoTech, line 148: Possible bug... Earlier in the method there is an exact-type fast path:

if (value is T returnValue)
{
    return returnValue;
}

That helps for As<Guid>(), but not for As<Guid?>() when value is a native Guid. In that case:

value is T is false because Guid is not Guid?
targetType becomes Guid
the Guid branch only accepts string
it throws
So I think this likely breaks:

- ((object)guid).As<Guid?>()
- any nullable-guid conversion path through IRecord.Get<Guid?>, TryGet<Guid?>, mapping, etc.

Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamReader.cs
AndyHeap-NeoTech: Minor thing, noticed that this is using Stream.ReadByte() whereas other methods are using NextByte(). Might be worth checking the consistency.

AndyHeap-NeoTech: Same use of Stream.ReadByte() as above where other methods are using NextByte().

AndyHeap-NeoTech: Same use of Stream.ReadByte()

AndyHeap-NeoTech (changes requested):
I'm not sure I saw tests for this kind of thing:

record.TryGet<Guid>("id", out var value)
record.GetCaseInsensitive<Guid>("ID")
nullable variants if supported/expected

I think there might be a slight bug, if you agree then adding the tests will help coverage and capture.

Neo4j.Driver/Neo4j.Driver/Public/Extensions/ValueExtensions.cs
RichardIrons-neo4j: Ooh, this is a chewy one, thanks! On it.

RichardIrons-neo4j: Turns out the fast path already handles `Guid?` because a boxed `Guid` does match `Guid?` at runtime. Added tests for TryGet, GetCaseInsensitive, and the nullable variants to demonstrate it explicitly.

Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamReader.cs
RichardIrons-neo4j: These are switched.

Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamReader.cs
```diff
@@ -142,12 +145,29 @@ public object ReadStruct()
         throw new ProtocolException("Unknown structure type: " + signature);
     }
 
+    public Guid ReadUuid()
+    {
+        if (Format.Version < BoltProtocolVersion.V6_1)
```
AndyHeap-NeoTech, line 150: Just a thought. Is this version check necessary. A protocol that is less than 6.1 will return an unknown type I think.

AndyHeap-NeoTech (approved):
One comment, but not a blocker.

Neo4j.Driver/Neo4j.Driver/Internal/IO/PackStreamReader.cs
RichardIrons-neo4j: Yeah - I'm beginning to use `ProtocolException`s as kind of assertions, or documented assumptions, at the transport level. The benefit is really more to me when I'm reading it.

PR #935, opened 2026-05-22
Fix MappingBindingsAttribute nullable value type properties

RichardIrons-neo4j:
Fixes DRIVERS-386

`MappingBindingsAttribute` declared some properties as nullable value types. C# doesn't allow `Nullable<T>` as a named attribute argument, so writing `[MappingBindings(Source = EntityMappingSource.NodeLabel)]` would get a compile error. The class itself compiled fine, and didn't have unit tests, so this went undetected.

I added tests that would fail (to even compile) for all the mapping attributes and changed the properties not to be nullable, with the nullability handled internally.


AndyHeap-NeoTech (approved):
Approved.

Issue #938, opened 2026-06-14
OverflowException when passing record types as parameters


TheJugglingOne:
**Describe the bug**
Passing a C# record instance as a query parameter causes an internal OverflowException during parameter serialization.

**To Reproduce**
```
using Neo4j.Driver;

public record TestDto
{
    public string Id { get; init; } = "123";
}

class Program
{
    static async Task Main()
    {
        var driver = GraphDatabase.Driver(
            "bolt://localhost:7687",
            AuthTokens.Basic("neo4j", "neo4j"));

        await driver.ExecutableQuery("RETURN 1")
            .WithParameters(new
            {
                test = new TestDto()
            })
            .ExecuteAsync();

        Console.WriteLine("Success");
    }
}
```

**Expected behavior**
Record instances should be serialized as parameter maps (similar to anonymous objects or dictionaries), with any unsupported or compiler-generated members safely ignored.

**Actual behavior**
The call fails with:
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.RuntimeTypeHandle.GetArgumentTypesFromFunctionPointer(QCallTypeHandle type, ObjectHandleOnStack argTypes)
   at System.RuntimeTypeHandle.GetArgumentTypesFromFunctionPointer(QCallTypeHandle type, ObjectHandleOnStack argTypes)
   at System.RuntimeTypeHandle.GetArgumentTypesFromFunctionPointer(RuntimeType type)
   at System.RuntimeType.RuntimeTypeCache.FunctionPointerCache.Create(RuntimeType type)
   at System.RuntimeType.IGenericCacheEntry`1.CreateAndCache(RuntimeType type)
   at System.RuntimeType.RuntimeTypeCache.get_FunctionPointerReturnAndParameterTypes()
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
   --- End of inner exception stack trace ---
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
[...]

PR #939, opened 2026-06-16
Fix OverflowException when passing record types as query parameters

RichardIrons-neo4j:
Fixes DRIVERS-452. Bug report: #938.

Failing tests were written for the reported bug (record types) and three other property kinds that cause their own exceptions: write-only, indexers, and static properties.

The fix filters the properties used to build the parameter dictionary to only public, readable, non-indexed instance properties.


AndyHeap-NeoTech (approved):
LGTM.

Issue #937, opened 2026-06-13
Connection acquisition timeout does not interrupt stale idle connection liveness probe


wallism:
I’m seeing intermittent user-visible stalls/errors in a production .NET app using the Neo4j .NET driver against AuraDB. I’m opening this issue to share evidence, check whether others are seeing the same pattern, and link it to a PR I’m working on to improve the behavior.

## Environment

- Neo4j.Driver: production package 6.1.1; investigated against 6.1.2 source
- Runtime: .NET 10
- Database: Neo4j AuraDB over Bolt/TLS
- Hosting: Google Cloud Run
- Traffic: low/moderate; issue is intermittent
- Relevant driver settings:
  - MaxConnectionPoolSize=75
  - MaxIdleConnectionPoolSize=10
  - MaxConnectionLifetime=10m
  - ConnectionIdleTimeout=2m
  - ConnectionLivenessCheckTimeout=5s
  - ConnectionAcquisitionTimeout=10s
  - TransactionTimeout=120s
- Server read timeout observed via driver logging: about 60s

## Summary

Instrumentation suggests a stale idle pooled connection can be selected for liveness probing, then block inside the liveness probe until the socket read timeout. The caller’s connection acquisition timeout fires earlier, but the underlying probe continues and the connection is only destroyed much later.

This creates user-visible request failure even though (spinning wheels):
- the pool is not saturated
- idle connections are available
- the Cypher query has not started
- adjacent graph operations complete quickly

## Observed Sequence

One production occurrence:

- Pool acquire starts for a write transaction.
- Pool snapshot: poolSize=13, idle=10, inUse=2, maxPool=75.
- Driver selects idle connection `bolt-65083`.
- Connection age/idle state means it requires liveness probing.
- Liveness probe starts via reset/sync.
- At T+10s, caller fails with:
  - `Failed to obtain a connection from pool within 00:00:10`
  - application transaction delegate had not started: `WorkStarted=False`
- At T+60s, the underlying socket read finally fails with `ConnectionReadTimeoutException`.
- Only then is the stale connection destroyed.
- Refresh/retry usually succeeds.

## Expected Behavior

If acquisition cancellation or `ConnectionAcquisitionTimeout` fires while an idle-connection liveness probe is blocked, the driver should actively abandon/destroy that candidate connection and continue/fail promptly.

Alternatively, the liveness probe should have its own short configurable timeout independent of the server socket read timeout.

## Actual Behavior

[...]

PR #940, opened 2026-06-16
Interrupt stale liveness probe on acquisition timeout (DRIVERS-454)

RichardIrons-neo4j:
Fixes DRIVERS-454, issue #937.

When the pool acquires an idle connection that needs a liveness probe, it runs `ResetAsync` + `SyncAsync`. If the socket is dead and the server never responds, `SyncAsync` blocks until the OS socket read timeout (~60s). The cancellation token was not passed into that read, so the connection-acquisition timeout only cancelled the wrapper — the blocking read kept running and held the pool slot until the OS timeout. The pool only releases the slot (decrements its connection count) when the connection is destroyed, so a hung probe makes the pool behave as if it were at MaxConnectionPoolSize, refusing new connections until the 60s timeout.

By threading the cancellation token all the way down the IO stack, we can actually interrupt the socket read and let the pool free the slot immediately.

The test `ShouldNotLeakPoolSizeWhenLivenessProbeHangsDuringAcquisitionTimeout` was written to isolate the bug before fixing. It simulates a hanging probe and asserts the connection is destroyed and the pool slot is released.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
@@ -219,13 +219,13 @@ internal sealed class SocketConnection : IConnection
         return AuthTokenManager.HandleSecurityExceptionAsync(AuthToken, exception);
     }
 
-    public async Task SyncAsync()
+    public async Task SyncAsync(CancellationToken cancellationToken = default)
     {
-        await SendAsync().ConfigureAwait(false);
-        await ReceiveAsync().ConfigureAwait(false);
+        await SendAsync(cancellationToken).ConfigureAwait(false);
+        await ReceiveAsync(cancellationToken).ConfigureAwait(false);
     }
 
-    public async Task SendAsync()
+    public async Task SendAsync(CancellationToken cancellationToken = default)
     {
         if (_messages.Count == 0)
             // nothing to send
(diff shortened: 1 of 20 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
```diff
@@ -219,25 +219,25 @@ public ValueTask<bool> NotifySecurityExceptionAsync(SecurityException exception)
...
+    public async Task SendAsync(CancellationToken cancellationToken = default)
     {
         if (_messages.Count == 0)
             // nothing to send
         {
             return;
         }
 
-        await _sendLock.WaitAsync().ConfigureAwait(false);
+        await _sendLock.WaitAsync(cancellationToken).ConfigureAwait(false);
```
AndyHeap-NeoTech, line 236: Doesn't this broadens cancellation semantics beyond the IO issue? Any shared code paths, not just liveness probing, could be affected. 

Not a bug, but a possible behaviour change I suppose.

AndyHeap-NeoTech (approved):
Looks good with one comment.

Neo4j.Driver/Neo4j.Driver/Internal/Connector/SocketConnection.cs
RichardIrons-neo4j: Just checked through the code paths. It's basically only affecting when connection acquisition requires a liveness probe.

PR #942, opened 2026-06-22
Support optional constructor and record parameters via default values

### LINKED ISSUE #854: MappingOptionalAttribute should be usable on record parameters?
**Describe the bug**
While `[MappingSource]` and `[MappingDefaultValue]` can be applied to record constructor params, the `[MappingOptional]` cannot. Is this intentional? Even its XML docs seem to suggest it was meant for properties or something else: "If a property or is decorated...".

Because of this, I can't just declare this on optional record constructors. Instead, I have to declare a separate property and add `[MappingOptional]` to it, which seems like unnecessary code.

**To Reproduce**
```csharp
public record MyObject(
  string Id,
  [MappingOptional] // <-- produces compiler error
  string? Extra = null
);
```

or

```csharp
public record MyObject(
  string Id,
  [property: MappingOptional] // <-- compiles, but default mapper doesn't recognize property as optional
  string? Extra = null
);
```

In the second example, it compiles, but runtime will fail with "Cannot map record to type Entity because the record does not contain a value for the parameter 'Extra'." if the property isn't present on the record.

**Expected behavior**
Should be able to just declare:

```csharp

public record MyObject(
  string Id,
  [MappingOptional]
  string? Extra = null
);
```

and default mapper will mark the `Extra` property as optional when reading records.

**Version Info**
 - .NET Version: net8.0
 - .NET Driver Version: 5.28.2
 - Neo4j Server Version & Edition: Desktop v2.0.1 (Windows)

RichardIrons-neo4j:
Depends on #943.

Fixes DRIVERS-462

Map absent record fields to constructor/record parameter defaults.

The object mapper previously threw whenever a record was missing a value for a constructor or record parameter, even when that parameter declared a default value. Parameters can now be made optional in the usual idiomatic way: 

```c#
record MyRecord(string Name, int? Age = null);

// or
class MyClass
{
    public MyClass(string key, int count = 0);
}
```

and the parameter mapper will honour the default value, with no need for attributes.

[MappingDefaultValue] continues to work as before, and [MappingOptional] stays property-only.

Resolves #854.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Mapping/OptionalMappingTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Mapping/OptionalMappingTests.cs
@@ -69,6 +69,45 @@ public class OptionalMappingTests
     }
 
     [Fact]
+    public void ShouldUseParameterDefaultWhenRecordParameterIsAbsent()
+    {
+        var record = TestRecord.Create(["Value"], [69]);
+        var mapped = record.AsObject<RecordWithDefaultParameter>();
+
+        mapped.Value.Should().Be(69);
+        mapped.Extra.Should().Be("implicit");
+    }
+
+    [Fact]
+    public void ShouldUseParameterDefaultWhenConstructorParameterIsAbsent()
+    {
+        var record = TestRecord.Create(["value"], [69]);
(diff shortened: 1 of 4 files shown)


AndyHeap-NeoTech (approved):
This looks good, nothing I can see to comment on.

PR #945, opened 2026-06-23
Add configurable connection read timeout cap

wallism:
## Summary
- add an optional `Config.ConnectionReadTimeoutCap` setting
- add `ConfigBuilder.WithConnectionReadTimeoutCap(TimeSpan)`
- cap server-provided `connection.recv_timeout_seconds` hints when the configured cap is lower

## Motivation
Servers can advertise a socket read timeout that is higher than an application's desired failure/retry window. This adds an opt-in upper bound while preserving the existing default behavior when no cap is configured.

## Tests
- `dotnet test Neo4j.Driver\Neo4j.Driver.Tests\Neo4j.Driver.Tests.csproj --filter "FullyQualifiedName~ConfigTests|FullyQualifiedName~HelloResponseHandlerTests"`

Note: a full unfiltered `Neo4j.Driver.Tests` run was started locally but exceeded my 3-minute command timeout, so the focused tests above are the verified local check.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConfigTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConfigTests.cs
@@ -40,6 +40,7 @@ public class ConfigTests
             config.Neo4JLogger.Should().BeOfType<NullNeo4JLogger>();
             config.MaxIdleConnectionPoolSize.Should().Be(100);
             config.ConnectionTimeout.Should().Be(TimeSpan.FromSeconds(30));
+            config.ConnectionReadTimeoutCap.Should().BeNull();
             config.TlsVersion.Should().Be(SslProtocols.Tls12);
             config.TlsNegotiator.Should().BeNull();
         }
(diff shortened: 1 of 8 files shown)


wallism:
This relates to an open support case for Aura: 00049079
I had cloned 6.1 locally and applied this timeout logic to that because otherwise my users need to wait 60s for existing timeouts. I had hoped 6.2 would address the issue but it does not and so I have to go back to my local modified clone so I can configure the underlying timeout to 10s.

RichardIrons-neo4j:
@wallism Hi. We discussed this in the team, with reference to your open Aura issue, and our decision was that the driver is not the right place to set this. Since the server will still send NOOPs at its configured rate an no faster, any long running query that exceeds the client’s set read-timeout will die, the connection will be killed and the server will be purged from the routing table. This is not a situation we would want to enable.

wallism:
Fair enough, that makes sense. Thanks for investigating.

PR #946, opened 2026-06-25
Add integration tests for native UUID type

RichardIrons-neo4j:
Adds integration tests for the native UUID type (Bolt 6.1, mapped to `System.Guid`), the driver-side counterpart to the testkit ITs.

`UuidIT` covers scalar echo (nil / max / sequential / random), lists, maps, storage on a node, and server-generated `uuid()`. Tests are gated to server `2026.5.0+`.

Verified locally against a 6.1 server — all 5 tests pass.

Closes DRIVERS-448.

> [!NOTE]
> While UUID is in preview, the integration-test server is now started with the required internal preview flags (version-gated, tagged `[uuid-preview]`), so CI actually negotiates 6.1. These workarounds are tracked for removal at GA in DRIVERS-476.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/StandAlone/ExternalBoltkitInstaller.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/StandAlone/ExternalBoltkitInstaller.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Neo4j.Driver.Internal;
+using Neo4j.Driver.Internal.Util;
 using Org.BouncyCastle.Pkcs;
 using static Neo4j.Driver.IntegrationTests.Internals.DefaultInstallation;
 using static Neo4j.Driver.IntegrationTests.Internals.SettingsHelper;
@@ -73,7 +74,30 @@ public class ExternalBoltkitInstaller : IInstaller
             DefaultConfig[ListenAddr] = Ipv6EnabledAddr;
         }
 
+        // TODO: remove once UUID / Bolt 6.1 is GA (DRIVERS-476)
+        // [uuid-preview] search tag for removal of UUID preview workarounds
+        if (ServerSupportsUuidPreview())
+        {
+            DefaultConfig["internal.dbms.bolt.max_protocol_version"] = "6.1";
+            DefaultConfig["internal.cypher.uuid_type_enabled"] = "true";
+            DefaultConfig["internal.dbms.latest_runtime_version"] = "2147483647";
+            DefaultConfig["internal.dbms.latest_kernel_version"] = "254";
+        }
+
         UpdateSettings(DefaultConfig);
+    }
+
+    // [uuid-preview] search tag for removal of UUID preview workarounds
+    private static bool ServerSupportsUuidPreview()
+    {
+        try
+        {
+            return ServerVersion.From(BoltkitHelper.ServerVersion()) >= ServerVersion.From("2026.5.0");
+        }
+        catch
+        {
+            return false;
+        }
     }
 
     public ISet<ISingleInstance> Start()
--- a/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
@@ -51,7 +51,7 @@ public sealed class UuidIT : DirectDriverTestBase
         {
             Guid.Empty,
             new("ffffffff-ffff-ffff-ffff-ffffffffffff"),
-            Guid.NewGuid(),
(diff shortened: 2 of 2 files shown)


Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/StandAlone/ExternalBoltkitInstaller.cs
robsdedude, line 1: I'm not seeing any test matrix being adjusted. Where is the code that decides against which server versions the tests are run?

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    [RequireServerFact(MinUuidServerVersion, GreaterThanOrEqualTo)]
+    public async Task ShouldSendAndReceiveInList()
+    {
+        var uuids = new List<Guid>
+        {
+            Guid.Empty,
+            new("ffffffff-ffff-ffff-ffff-ffffffffffff"),
+            Guid.NewGuid(),
```
robsdedude: Why is this not `new Guid("01020304-0506-0708-090a-0b0c0d0e0f12")`? Not that it really matters, I'm just curious.

PS: I realize in TestKit the list in this test only contains the first 3 elements.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
```diff
@@ -0,0 +1,150 @@
...
+    [RequireServerFact(MinUuidServerVersion, GreaterThanOrEqualTo)]
+    public async Task ShouldStoreAndRetrieveUuidOnNode()
+    {
+        var uuid = Guid.NewGuid();
+
+        var session = Server.Driver.AsyncSession(o => o.WithDefaultAccessMode(AccessMode.Write));
+        try
+        {
+            var cursor = await session.RunAsync(
+                "CREATE (n:Node { id: $uuid }) RETURN n.id",
```
robsdedude, line 104: Are all other tests resilient against an unclean DB state?

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    [RequireServerFact(MinUuidServerVersion, GreaterThanOrEqualTo)]
+    public async Task ShouldReceiveServerGeneratedUuid()
+    {
+        var session = Server.Driver.AsyncSession(o => o.WithDefaultAccessMode(AccessMode.Read));
+        try
+        {
+            var cursor = await session.RunAsync("RETURN uuid()");
```
robsdedude, line 123: Along the lines of https://github.com/neo4j-drivers/testkit/pull/723/changes#diff-957c6f7d0342e4c657c63896ea9938572e597f80434015145aa905271eae9714R57

Wouldn't it be better to generate a non-random UUID server side and to guarantee that the driver produces the right value? Being able to round-tip alone isn't enough to know that server and driver agree on the encoding.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
RichardIrons-neo4j: I agree that it's better, I put it in.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Internals/StandAlone/ExternalBoltkitInstaller.cs
RichardIrons-neo4j: It's not in this repo:

- The Tests.Integration suite doesn't define the matrix. Each run gets one server version; RequireServerFact(version, …) just skips if the running server doesn't satisfy it.
- The version arrives via runTests.ps1 (-ServerVersion → NEOCTRL_ARGS).
- The list of versions lives in our TeamCity config (one build per version), not the repo.
- These tests start running once a 2026.5+ server is in that set — I'll make sure it's covered.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
RichardIrons-neo4j: The test base class, DirectDriverTestBase wipes the db on Dispose, so yes.

Neo4j.Driver/Neo4j.Driver.Tests.Integration/Types/UuidIT.cs
RichardIrons-neo4j: Good spot - added a new test for this.

robsdedude (approved):
🥼

PR #949, opened 2026-06-30
Move .NET-specific testkit skip decisions into the test backend

RichardIrons-neo4j:
Fixes DRIVERS-480

## Summary

Moves driver-specific test-skip decisions out of the shared testkit and into the .NET test backend, using testkit's existing `StartTest`/`StartSubTest` protocol. This keeps the shared testkit driver-agnostic and lets the interim `get_driver_name() not in ["dotnet"]` guards be removed (testkit side, separate change).

The motivating case: testkit's `test_zoned_time` includes UTC offsets of ±86400 (±24h) and ±64920 (±18h02m), which exceed .NET `OffsetTime`'s ±18:00 (±64800s) limit and can't be constructed. The backend now decides, per subtest, whether a value is representable.

## What's here

A small policy abstraction under `Protocol/Skip/`:

- `TestDisposition` (`RunAll` / `RunSubtests` / `SkipAll(reason)`) as a sealed-record union, plus a `TestDispositions` factory for enum-like values.
- `ITestSkipPolicy` (`Overall` + `ShouldRunSubtest(parameters, out reason)`), with `RunAllPolicy` (default), `SkipAllPolicy`, and `TryConstructPolicy`.
- `TestSkipPolicyRegistry` keyed by test name (substring match), with `TestSkipPolicies.Default` unifying the existing whole-test blacklist (`TestBlackList`, folded in as `SkipAll`) and per-subtest policies.
- `StartTest` now routes through the registry; a new `StartSubTest` handler maps `ShouldRunSubtest` to `RunTest`/`SkipTest`.
- `TryConstructPolicy` attempts to convert each subtest argument to its native value via the existing `CypherToNative` path; any construction failure means the value can't be represented, so it's skipped with the real exception message. Registered for `test_zoned_time`.

## Backend hardening (also fixes a latent cascade bug)

`CypherToNative.Convert` no longer wraps conversion failures in `IOException`. The controller treats `IOException` as a socket-level failure and tears down the connection, so a single unrepresentable value was killing the connection and cascading the remaining subtests into `BrokenPipe`. Conversion now runs outside any catch, so the real exception (an `ArgumentException`) reaches the controller's clean per-request handler; unknown/unmapped types throw `ArgumentException` rather than `IOException`.

The baseline commit also brings the DRIVERS-117 test-backend wire (de)serialization (hex bytes, NaN/±Infinity) and a new `TestBackend.Tests` xUnit project across.

## Tests

28 backend unit tests (xUnit + FluentAssertions), covering the dispositions/policies, the registry + blacklist fold-in, and the full `StartTest`/`StartSubTest` wiring end-to-end (out-of-range zoned-time offset → skip with the real reason; in-range → run).

## Notes for review

- The matching testkit changes (revert the interim skip, reshape `test_zoned_time` to use `should_run_subtest`) are a separate change on the testkit `feat/http-query-api` branch.
- The `Feature:HTTP:QueryAPI:1.0` flag is intentionally **not** added here (it isn't true on `6.x`), so the `http_query` suites skip on this branch; the mechanism is verified by the backend unit tests.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend.Tests/NativeToCypherTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests.TestBackend.Tests/NativeToCypherTests.cs
@@ -1,0 +1,35 @@
+using System;
+using FluentAssertions;
+using Neo4j.Driver.Tests.TestBackend.Types;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Neo4j.Driver.Tests.TestBackend.Tests;
+
+public class NativeToCypherTests
+{
+    [Fact]
+    public void Convert_Bytes_ProducesSpaceSeparatedHex()
+    {
+        var bytes = new byte[] { 0x00, 0x33, 0x66, 0x99, 0xcc, 0xff };
+
+        var result = (NativeToCypherObject)NativeToCypher.Convert(bytes);
+
+        result.name.Should().Be("CypherBytes");
+        ((NativeToCypherObject.DataType)result.data).value.Should().Be("00 33 66 99 cc ff");
+    }
+
+    [Theory]
+    [InlineData(double.NaN, "NaN")]
+    [InlineData(double.PositiveInfinity, "+Infinity")]
+    [InlineData(double.NegativeInfinity, "-Infinity")]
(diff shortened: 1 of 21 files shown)


RichardIrons-neo4j (commented):
Self-approving since it's test backend code

PR #952, opened 2026-07-07
Add MEL-shaped internal logging stack

RichardIrons-neo4j:
## Summary

Brings the Microsoft.Extensions.Logging (MEL)-shaped internal logging infrastructure the Query API work uses into the driver, as self-contained, dormant code.

Depends on #953.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
@@ -45,7 +45,14 @@ internal class ContextualLogger : ILogger
             return;
         }
 
-        using var scope = _downstream.BeginScope(_tracker.Contexts.AsMicrosoftStateItems().ToList());
+        var contexts = _tracker.Contexts;
+        if (contexts.Count == 0)
+        {
+            _downstream.Log(logLevel, eventId, state, exception, formatter);
+            return;
+        }
+
+        using var scope = _downstream.BeginScope(contexts.AsMicrosoftStateItems().ToList());
         _downstream.Log(logLevel, eventId, state, exception, formatter);
     }
 
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#nullable enable
+
 using System;
 using System.Linq;
 
@@ -31,7 +33,12 @@ internal class ContextualLogger : ILogger
         _downstream = downstream;
     }
 
-    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+    public void Log<TState>(
+        LogLevel logLevel,
+        EventId eventId,
+        TState state,
+        Exception? exception,
+        Func<TState, Exception?, string> formatter)
     {
         using var scope = _downstream.BeginScope(_tracker.Contexts.AsMicrosoftStateItems().ToList());
         _downstream.Log(logLevel, eventId, state, exception, formatter);
@@ -42,9 +49,8 @@ internal class ContextualLogger : ILogger
         return _downstream.IsEnabled(logLevel);
     }
 
-    public IDisposable BeginScope<TState>(TState state) where TState : notnull
+    public IDisposable? BeginScope<TState>(TState state) where TState : notnull
     {
         return _downstream.BeginScope(state);
     }
 }
-
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) "Neo4j"
+// Neo4j Sweden AB [https://neo4j.com]
+// 
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Linq;
+
+namespace Neo4j.Driver.Internal;
+
+internal class ContextualLogger : ILogger
+{
+    private readonly ILoggingContextTracker _tracker;
+    private ILogger _downstream;
+
+    public ContextualLogger(
+        ILoggingContextTracker tracker,
+        ILogger downstream)
+    {
+        _tracker = tracker;
+        _downstream = downstream;
+    }
+
+    public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+    {
+        using var scope = _downstream.BeginScope(_tracker.Contexts.AsMicrosoftStateItems().ToList());
+        _downstream.Log(logLevel, eventId, state, exception, formatter);
+    }
+
+    public bool IsEnabled(LogLevel logLevel)
+    {
+        return _downstream.IsEnabled(logLevel);
+    }
+
+    public IDisposable BeginScope<TState>(TState state) where TState : notnull
+    {
+        return _downstream.BeginScope(state);
+    }
+}
+
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
@@ -40,6 +40,11 @@ internal class ContextualLogger : ILogger
         Exception? exception,
         Func<TState, Exception?, string> formatter)
     {
+        if (!_downstream.IsEnabled(logLevel))
+        {
+            return;
+        }
+
         using var scope = _downstream.BeginScope(_tracker.Contexts.AsMicrosoftStateItems().ToList());
         _downstream.Log(logLevel, eventId, state, exception, formatter);
     }
--- a/Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
+++ b/Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
@@ -16,10 +16,6 @@
 #nullable enable
 
 using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text.RegularExpressions;
 
 namespace Neo4j.Driver.Internal;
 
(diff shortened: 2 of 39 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) "Neo4j"
```
RichardIrons-neo4j, line 1: Most of the code in this file is just recreating the shape of the Microsoft.Extensions.Logging framework, and will eventually disappear when we add a dependency on that.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
```diff
@@ -0,0 +1,191 @@
...
+    Debug = 1,
+    Information = 2,
+    Warning = 3,
+    Error = 4,
+    Critical = 5,
+}
+
+internal record EventId(int Id, string Name);
+
+internal static class LoggerExtensions
```
AndyHeap-NeoTech, line 48: Thinking a class of this size should be in it's own file.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
```diff
@@ -0,0 +1,191 @@
...
+        => logger.Log(logLevel, new EventId(0, ""), exception, message, args);
+
+    public static void Log(this ILogger logger, LogLevel logLevel, EventId eventId, Exception? exception, string? message, params object?[] args)
+    {
+        ArgumentNullException.ThrowIfNull(logger);
+        logger.Log(logLevel, eventId, new LogParams(message ?? "", args), exception, DefaultFormatter);
+    }
+}
+
+internal class LogParams : IReadOnlyList<KeyValuePair<string, object?>>
```
AndyHeap-NeoTech: And possibly this class in it's own file as well.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
```diff
@@ -0,0 +1,191 @@
...
+        {
+            result.Add(new(parameterNames[i], args[i]));
+        }
+
+        return result;
+    }
+
+    private static List<string> ExtractParameterNames(string messageTemplate)
+    {
+        var matches = Regex.Matches(messageTemplate, @"\{(\w+)\}");
```
AndyHeap-NeoTech: This one is something claude code found:

Format/alignment specifiers desync the two regexes → FormatException at log time.
LogParams.ExtractParameterNames uses \{(\w+)\}, but LegacyLoggerAdapter.GetModifiedFormat uses \{[^}]+\}. They only agree when every placeholder is a bare word. Trace it for LogDebug("took {ms:D3}ms", 42):
- \{(\w+)\} does not match {ms:D3} (colon breaks it) → LogParams captures zero params → args extracted = [].
- \{[^}]+\} does match it → template becomes "[Type] took {0}ms".
- Result: _legacyLogger.Debug("[Type] took {0}ms", new object[0]) → String.Format throws FormatException.

MEL templates legitimately use {count:D3} / {value,10}, so this is a real trap. The two placeholder parsers must share one regex/one source of truth.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
```diff
@@ -0,0 +1,58 @@
...
+        [NotNullWhen(true)] out object?[]? args)
+    {
+        if (state is not IEnumerable<KeyValuePair<string, object>> list)
+        {
+            format = null;
+            args = null;
+            return false;
+        }
+
+        var dict = list.ToDictionary(kv => kv.Key, object? (kv) => kv.Value);
```
AndyHeap-NeoTech: Duplicate placeholder names throw ArgumentException at log time.
Doing this list.ToDictionary(kv => kv.Key, ...) when LogParams happily emits two entries with the same key for "{x}={x}", so ToDictionary throws on the duplicate. It also relies on Dictionary<,> preserving insertion order for argument ordering, which isn't guaranteed.

Might be better to iterate the list directly.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Text.RegularExpressions;
+using Neo4j.Driver.Internal.Util;
+
+namespace Neo4j.Driver.Internal;
+
+internal class LegacyLoggerAdapter : ILogger
+{
+    private readonly INeo4jLogger _legacyLogger;
+    private readonly Type _loggingType;
+    private string _scopePrefix = "";
```
AndyHeap-NeoTech: Another that Claude Code flagged:

Thread-safety: the scope mechanism is a mutable instance field.
LegacyLoggerAdapter._scopePrefix is a plain mutable field. ContextualLogger.Log sets it via BeginScope, logs, then restores on dispose — fine single-threaded, but a driver runs many sessions/transactions concurrently. If one ILogger instance for a type is shared across threads (the normal expectation for loggers, and what a per-type DI cache would produce), two concurrent Log calls race on _scopePrefix and emit cross-contaminated context prefixes. Real MEL implementations keep scope state in AsyncLocal, precisely for this. Same concern applies to LoggingContextTracker._contexts (plain List, Add/Remove/enumerate with no synchronisation → InvalidOperationException if walked while mutated). Worth deciding the intended lifetime/threading contract now and encoding it, because it can't be retrofitted invisibly later.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
```diff
@@ -0,0 +1,191 @@
...
+    public static void Log(this ILogger logger, LogLevel logLevel, EventId eventId, string? message, params object?[] args)
+        => logger.Log(logLevel, eventId, null, message, args);
+
+    public static void Log(this ILogger logger, LogLevel logLevel, Exception? exception, string? message, params object?[] args)
+        => logger.Log(logLevel, new EventId(0, ""), exception, message, args);
+
+    public static void Log(this ILogger logger, LogLevel logLevel, EventId eventId, Exception? exception, string? message, params object?[] args)
+    {
+        ArgumentNullException.ThrowIfNull(logger);
+        logger.Log(logLevel, eventId, new LogParams(message ?? "", args), exception, DefaultFormatter);
```
AndyHeap-NeoTech: new logParams runs without a check for IsEnabled. Means there is a cost even when logging is disabled and could cause issues on hot paths.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public void Log<TState>(
+        LogLevel logLevel,
+        EventId eventId,
+        TState state,
+        Exception? exception,
+        Func<TState, Exception?, string> formatter)
+    {
+        using var scope = _downstream.BeginScope(_tracker.Contexts.AsMicrosoftStateItems().ToList());
```
AndyHeap-NeoTech: I'm thinking an IsEnabled check here would be a good idea to prevent unnecessary work on hot paths.

AndyHeap-NeoTech: This line will overwrite anything set with the BeginScope method call. If direct BeginScope use isn't intended then it's ok, but that method is a public interface on the class.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Neo4j.Driver.Internal;
+
+internal static class LoggingHelpers
+{
+    private const string OriginalFormatStringKey = "{OriginalFormat}";
+
+    public static bool TryBuildScopePrefix<TState>(TState state, [NotNullWhen(true)] out string? prefix)
+        where TState : notnull
+    {
+        if (state is not IEnumerable<KeyValuePair<string, object?>> contexts)
```
AndyHeap-NeoTech: This method, TryBuildScopePrefix, and ExtractFormatAndArguments both match IEnumerable<KeyValuePair<string, object?>> - they're distinguished only by which call site feeds them (LogParams vs the contexts list), not by type. A future refactor that crosses those wires would fail silently.  Consider a marker type for LogParams.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
RichardIrons-neo4j: Ah, the thing about this, is this whole file is stuff that will get deleted as soon as we reference Microsoft.Extensions.Logging. Most of it is literally copied from their source. So it's just all bundled together because it's not "real" code.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
RichardIrons-neo4j: Split LogParams out into its own file.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
RichardIrons-neo4j: Added failing test and fixed - both sides now share one regex, and specifiers are preserved ({ms:D3} -> {0:D3}).

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
RichardIrons-neo4j: Added test and fixed - iterates the list directly now.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
RichardIrons-neo4j: Went with AsyncLocal for the scope prefix; also synchronised the tracker's context list.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ILogger.cs
RichardIrons-neo4j: Added an IsEnabled guard.

Neo4j.Driver/Neo4j.Driver/Internal/Core/ContextualLogger.cs
RichardIrons-neo4j: Added.

RichardIrons-neo4j: It's only public on an internal type, so leaving as is.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
RichardIrons-neo4j: Done - ExtractFormatAndArguments now only matches LogParams, and TryBuildScopePrefix rejects it.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
```diff
@@ -0,0 +1,71 @@
...
+        where TState : notnull
+    {
+        // LogParams is message state, not scope context, even though it has the same enumerable shape.
+        if (state is LogParams || state is not IEnumerable<KeyValuePair<string, object?>> contexts)
+        {
+            prefix = null;
+            return false;
+        }
+
+        prefix = string.Concat(contexts.Select(kvp => $"[{kvp.Key}:{kvp.Value}] "));
```
AndyHeap-NeoTech, line 37: Scope context values are interpolated into the message template, not passed as format arguments - so a brace in any context value throws at log time.

The prefix built here ([key:value] ) is concatenated into the format string over in LegacyLoggerAdapter.GetModifiedFormat: 
`var finalFormat = $"[{typeName}] {_scopePrefix.Value}{indexedFormat}";`
and finalFormat is then handed to _legacyLogger.Debug/Info/..., which runs it through String.Format. Any { or } in a context value becomes part of the format string.

Session/tx ids are safe today, but nothing constrains a future context value (db name, query fragment, etc.) to be brace-free, and a logging shim shouldn't be able to throw into the calling path.

One option - double any braces when building the prefix, since it's fixed literal text by the time it reaches String.Format. 
`prefix = string.Concat(contexts.Select(kvp => $"[{kvp.Key}:{kvp.Value}] ".Replace("{", "{{").Replace("}", "}}")));`

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
```diff
@@ -0,0 +1,116 @@
...
+        _legacyLogger = legacyLogger;
+        _loggingType = loggingType;
+    }
+
+    private string GetModifiedFormat(string messageTemplate)
+    {
+        // Replace "{id}, {name:D3}" with "{0}, {1:D3}", preserving alignment/format suffixes.
+        // Must use the same placeholder definition as LogParams so args stay paired with placeholders.
+        var index = 0;
+        var indexedFormat = LogParams.PlaceholderRegex.Replace(
```
AndyHeap-NeoTech: When a caller passes fewer args than there are placeholders, this throws FormatException - and it's the same "two sources of truth" shape as the regex bug that is fixed.

It's a caller error to under-supply args, but a logging shim throwing into the calling path is the failure mode that ain't good.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LoggingHelpers.cs
RichardIrons-neo4j: Fixed

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
RichardIrons-neo4j: Fixed

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
```diff
@@ -0,0 +1,155 @@
...
+        TState state,
+        Exception? exception,
+        Func<TState, Exception?, string> formatter)
+    {
+        if(!IsEnabled(logLevel))
+        {
+            return;
+        }
+
+        if (!LoggingHelpers.ExtractFormatAndArguments(state, out var format, out var args))
```
AndyHeap-NeoTech: Log<TState> returns here for any state that isn't a LogParams, so the formatter argument is never used. That silently drops any log call not routed through LoggerExtensions and diverges from the MEL contract, where formatter is the fallback for arbitrary state. Safe today since every caller builds a LogParams, but worth either honouring formatter or documenting the LogParams-only contract.

AndyHeap-NeoTech (approved):
Just one comment more.

Neo4j.Driver/Neo4j.Driver/Internal/Core/LegacyLoggerAdapter.cs
RichardIrons-neo4j: Thanks - this would have been hard to debug if it ever happened. Put a fix in.

PR #955, opened 2026-07-16
Fix hanging liveness probe consuming acquisition timeout (DRIVERS-488)

RichardIrons-neo4j:
Fixes DRIVERS-488.

If a liveness check timeout is set, then the driver attempts to "ping" the server when it takes the connection from the pool, if the connection has been idle longer than the threshold. 

If in the meantime, the server has dropped the connection without the client getting a RST message, then the "ping" (actually a RESET) hangs. The only timeout guarding this is the ConnectionAcquisitionTimeout, typically set at 60s, which is "the longest I will wait to get a usable connection". Therefore because of the hung read, the entire timeout gets used up waiting for a response that never comes, so instead of trying a different connection, the whole acquire operation fails. 

The bug was reproduced as a failing test, with the fix turning it green.

The fix I've put in is to guard the liveness check by using the ConnectionTimeout, which is the timeout used when creating a new connection ("the longest I will try one connection before moving on").

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
@@ -220,6 +220,58 @@ public class ConnectionPoolTests
         }
 
         [Fact]
+        public async Task ShouldAcquireFreshConnectionWhenIdleConnectionLivenessProbeHangs()
+        {
+            var deadConnection = new Mock<IPooledConnection>();
+            deadConnection
+                .Setup(x => x.ResetAsync(It.IsAny<CancellationToken>()))
+                .Returns(Task.CompletedTask);
+
+            deadConnection
+                .Setup(x => x.SyncAsync(It.IsAny<CancellationToken>()))
+                .Returns<CancellationToken>(ct => Task.Delay(Timeout.Infinite, ct));
+
+            var freshConnection = new Mock<IPooledConnection>();
+            freshConnection
+                .Setup(x => x.InitAsync(It.IsAny<SessionConfig>(), It.IsAny<CancellationToken>()))
+                .Returns(Task.CompletedTask);
+
+            freshConnection
+                .Setup(x => x.ResetAsync(It.IsAny<CancellationToken>()))
+                .Returns(Task.CompletedTask);
+
+            freshConnection
+                .Setup(x => x.SyncAsync(It.IsAny<CancellationToken>()))
+                .Returns(Task.CompletedTask);
+
+            freshConnection
+                .Setup(x => x.ValidateCredsAsync())
+                .Returns(ValueTask.CompletedTask);
+
+            var factoryMock = new Mock<IPooledConnectionFactory>();
+            factoryMock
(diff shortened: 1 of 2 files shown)


Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
AndyHeap-NeoTech, line 1: Might be worth adding testing for the aquisitionTimeout wins situation to ensure the behaviour is as expected and doesn't change by accident in the future.

In the existing test the probe timeout fires first, the outer token is not cancelled, so ThrowIfCancellationRequested() is a no-op and the loop recovers to a fresh connection. In the inverted test the acquisition token fires first, it's linked into probeTimeoutSource so the probe still unblocks, but now ThrowIfCancellationRequested() should throw and the whole AcquireAsync should surface the acquisition-timeout failure (a ClientException from the .Timeout(...) wrapper in AcquireOrTimeoutAsync) rather than returning a connection.

AndyHeap-NeoTech (approved):
Approving, but want to note that DriverContext.Config.ConnectionTimeout can be set to infinite, which means the bug would resurface. Not a blocker as it needs to be opted into but thought I'd mention it incase you want to change anything because of it.

Also a possible test case that can be added.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -107,6 +107,7 @@ internal ConnectionPool(
 
     private int MaxIdlePoolSize => DriverContext.Config.MaxIdleConnectionPoolSize;
     private TimeSpan ConnectionAcquisitionTimeout => DriverContext.Config.ConnectionAcquisitionTimeout;
+    private TimeSpan ConnectionTimeout => DriverContext.Config.ConnectionTimeout;
```
AndyHeap-NeoTech, line 110: Had a chat with Max about this as I asked him to check the JS driver for the same bug. We're not sure that reusing ConnectionTimeout is the right call. Will discuss in stand-up next time.

Neo4j.Driver/Neo4j.Driver.Tests/ConnectionPoolTests.cs
RichardIrons-neo4j: Good idea about the inverse test, thanks. I've added that.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
RichardIrons-neo4j: Discussed with Max this morning, and we decided that we'll use ConnectionTimeout as a first step, and then when we're unifying the behaviour across drivers, introduce a new config option.

RichardIrons-neo4j:
> Approving, but want to note that DriverContext.Config.ConnectionTimeout can be set to infinite, which means the bug would resurface. Not a blocker as it needs to be opted into but thought I'd mention it incase you want to change anything because of it.
> 
> Also a possible test case that can be added.

Test added, thanks.

When they set ConnectionTimeout to infinite, they're basically saying that the Acquisition timeout is the only guard on anything, so I'm fine with this being included in that.

PR #957, opened 2026-08-05
Testkit backend rewrite

RichardIrons-neo4j:
Closes DRIVERS-512

This PR replaces the legacy testkit backend (`Neo4j.Driver.Tests.TestBackend.Legacy`) with a ground-up rewrite, `Neo4j.Driver.TestKitBackend`.

The main reason for the rewrite is cost of change. The testkit protocol grows with every driver feature, so the backend gets edited more often than most of the repo, and in the legacy backend each of those edits meant several files of hand-maintained plumbing. The rewrite is designed around making that recurring edit small.

## What a new message costs

A protocol message is one file: a request record, a response record, and a handler. This is a real handler from the rewrite, minus one debug log line:

```csharp
internal record CheckMultiDBSupportRequest : IProtocolMessage
{
    [StoredObject]
    public required IDriver Driver { get; init; }
    public required string DriverId { get; init; }
}

internal record MultiDBSupportResponse(string Id, bool Available) : IProtocolMessage;

internal class CheckMultiDBSupportHandler : MessageHandler<CheckMultiDBSupportRequest>
{
    private readonly IResponseWriter _responseWriter;

    public CheckMultiDBSupportHandler(IResponseWriter responseWriter)
    {
        _responseWriter = responseWriter;
    }

    public override async Task ProcessAsync(CheckMultiDBSupportRequest message)
    {
        var available = await message.Driver.SupportsMultiDbAsync();
        await _responseWriter.WriteAsync(new MultiDBSupportResponse(message.DriverId, available));
    }
}
```

There is nothing else to write. Registration is an assembly scan, so no list is edited. The record's properties are the wire format, so the protocol shape is visible in one place, and the record declares the shape it needs from the per-test object store: the one `driverId` wire field arrives as both the live `IDriver` (resolved during deserialization via `[StoredObject]`) and the id string the response echoes. Error handling lives in the read loop, so the handler catches nothing.

Multi-message flows stay close to this size. A driver callback (resolver, auth manager, bookmark manager, certificate provider) is three pieces: an outbound record, a one-line await, and a completion handler that fulfils the awaited expectation. Retryable transactions use the same mechanism keyed on the session id, and the flow reads as one linear method with the driver's own retry logic deciding what happens next. The backend's [README](Neo4j.Driver/Neo4j.Driver.TestKitBackend/README.md) builds each of these up step by step, from `NewDriver` to the retryable flow, with sequence diagrams.

Handlers are constructor-injected classes, which is what makes them unit-testable: the rewrite has generous coverage over handler and mapper logic.

## Where the complexity lives

The small handlers are paid for by deliberate machinery underneath: a polymorphic envelope converter for the `{"name","data"}` frame, attribute-driven handle resolution (`[StoredObject]`) and payload envelopes (`[ProtocolEnvelope]`), a correlation wrapper (`CorrelatedRequestWrapper`) that injects callback ids on the wire, and a keyed expectation store that lets a parked handler await a message that has not arrived yet. Each piece exists so that handlers never deal with JSON, ids, registration, or error frames.

The same structure keeps the machinery itself workable: each of those mechanisms is one converter or attribute plus its tests, messages opt in declaratively, and none of them is woven through the handlers. When a serialization concern changes, the change lands in the mechanism, not across the message files.

## Problems with the legacy backend

- Difficult to read and understand: every message is tripled (a class, a nested `data` DTO, and often a hand-maintained converter), and simple values get wrapped in chains like `result.data.auth.data.auth.data` (see `NewBearerAuthTokenManager.cs:44`).
- No real unit test coverage. The ~6,900 lines of protocol and handler code have 26 tests total, and none of them touch the actual message handlers.
- The backend logs an `IOException` at the end of every test, because it treats a normal connection close as a socket error.
- Difficult to change and maintain: adding one message type touches several files, and the retry flow works through recursive controller re-entry that a stray event can break.
- The backend never releases driver and session handles: they accumulate for the lifetime of the process in a global registry.
- Produced enormous log files. Nothing caps how much a single request or response can write, so one large-payload test can dump megabytes into the log.
- The backend itself hardcodes test skips: 40 blacklist entries baked directly into `TestBlackList.cs`, several with reasons no more specific than "Behaves strange" or "requires investigation." The rewrite audited this list rather than porting it: 25 entries were stale and are gone (those tests now run and pass), and the remainder carry specific, verified reasons, with the two genuine driver bugs tracked as DRIVERS-515 and DRIVERS-516.

## Architecture

- One record per message plus a DI-resolved handler, in the same file.
- A single polymorphic envelope serializer (System.Text.Json) for the `{"name","data"}` wire shape, with `Optional<T>` only where absent, null, and value need to stay distinct.
- Kestrel `ConnectionHandler` transport with one DI scope per connection. Testkit opens one connection per test, so scope disposal cleans up drivers and sessions automatically. No state leaks between tests.
- A continuously reading message loop that dispatches each message on its own task. Flows that span several messages (driver callbacks, retryable transactions) await a keyed expectation that the later message's handler fulfils, so each flow reads as one linear method.
- Fully async throughout, with no sync-over-async.
- Microsoft.Extensions.Logging backed by Serilog, configured from `appsettings`, with a formatter that caps line length so a multi-megabyte test payload can't blow up the log file.

## Correctness fixes

Some legacy behaviour turned out to be broken, not just structurally awkward, and is fixed here rather than ported:

- The address resolver now sends the address actually being resolved. Legacy always sent the driver's original URI, which only worked in the simple case where rediscovery re-resolves the initial router. `test_should_use_resolver_during_rediscovery_when_existing_routers_fail` used to hang for 60 seconds against legacy; it passes immediately here.
- Legacy's message loop could deadlock: operations ran synchronously on the loop up to their first incomplete `await`, so a callback seam in that prefix (such as the resolver) could block the loop before its own callback request was even written. The rewrite dispatches every message on its own task, so no handler can block the loop.
- The rewrite's retryable-negative design fixes the flaw behind the "Behaves strange" blacklist entries on two of testkit's clustering retry tests.
- The legacy backend treated a normal connection close as a socket error and logged an `IOException` after every test. An early version of this rewrite had the inverse flaw: a graceful close (EOF rather than an exception) left the read loop spinning on a pinned thread and kept the connection's DI scope alive. The loop now treats EOF, `IOException`, and unexpected read failures as the same clean shutdown path, cancelling outstanding expectations and disposing the scope.

## Test results and log size, before and after

Same testkit suite (6,412 tests), one run per backend:

| | Passed | Skipped | Failed |
|---|---|---|---|
| Legacy backend (last green `6.x` nightly, 2026-08-11) | 5,368 | 1,044 | 0 |
| Rewrite backend (this PR, head `ea2b3561`) | 5,576 | 836 | 0 |

The 208-test swing matches the skip-list audit above: tests hardcoded-skipped against the legacy backend now run and pass.

Backend log size, same artifact (`driver_backend/out.log`) from each run:

| | Size |
|---|---|
| Legacy backend | 722,057,295 bytes (≈ 689 MiB) |
| Rewrite backend | 124,170,374 bytes (≈ 118 MiB) |

An 83% reduction, from the truncating log formatter and console-by-default logging described under Architecture.

## Deliberate behavioural deltas from legacy

A few differences from the legacy backend are worth calling out explicitly, so they read as deliberate rather than overlooked.

- **`EncryptionManager` and `+s`/`+ssc` schemes.** This isn't a backend-specific delta: `EncryptionManager.Create` is shared core-driver code, used identically by both backends. While working on this PR we changed its conflict rule. It used to throw on *any* explicit `encrypted`/trust config combined with a `+s`/`+ssc` scheme. Now, security config explicitly set to the driver's defaults (`encrypted: false`, default chain trust) is treated as unset and the scheme wins silently, while any non-default explicit config still throws, *including* config that agrees with what the scheme implies (`encrypted: true` with `+s`). That asymmetry is testkit's cross-driver contract, not an accident: `tests/tls/test_explicit_options.py::test_explicit_config_and_scheme_config` requires an error for every non-default explicit security config combined with a secure scheme (java and javascript assert the same), and the `Detail:DefaultSecurityConfigValueEquality` feature flag is the license for the set-to-default carve-out. One side effect: a caller who explicitly passes default security config alongside a `+s`/`+ssc` URI gets no error and no signal that the scheme overrode them. On the encrypted flag this is harmless (it can only resolve to the scheme's own encrypted setting), but on the trust axis it is a real behavioural downgrade path: an explicitly configured default `ChainTrustManager` combined with `+ssc` now silently resolves to the accept-all `InsecureTrustManager`, where prior releases threw `ArgumentException`. The caller asked for certificate verification by name and gets none, with no signal. That is what the testkit contract requires (`+ssc` *means* no verification, and default-valued config counts as unset), but it is the strongest argument for a warning log on the silent-override path. We considered adding one and are deliberately not doing so in this PR; noting it here as a known, accepted gap rather than a regression.

- **Three new feature flags with no legacy equivalent**: `Backend:RTFetch` and `Backend:RTForceUpdate` (gating `GetRoutingTable`/`ForcedRoutingTableUpdate`), and `Feature:API:Type.Spatial` (gating point/spatial Cypher value handling). These are new CI surface; testkit stub tests behind these flags didn't run against the legacy backend at all.

- **`RunSubTests`/`StartSubTest` capability removed.** Legacy supported testkit's parameterized-subtest looping: a stub test could enumerate a matrix of argument combinations, and the backend would be asked, per combination, whether to run or skip it via a skip policy keyed on `testName` + `subtestArguments`. The rewrite doesn't implement this: no message type, no handler. This degrades safely rather than breaking. Testkit only sends `StartSubTest` if the backend opts in by replying `RunSubTests` to `StartTest`, and since ours never does, `self.subTest(...)` in every stub test falls back to Python-side bookkeeping with no wire message at all. The capability we lose is per-subtest granularity: a test with subtests now runs or skips as one atomic unit off the overall `StartTest` decision, instead of being skippable case-by-case.


Neo4j.Driver/Neo4j.Driver/Internal/EncryptionManager.cs
```diff
@@ -52,14 +62,14 @@ private static EncryptionManager CreateFromUriScheme(Uri uri, INeo4jLogger neo4J
...
-    private static void AssertSimpleUriScheme(Uri uri, EncryptionLevel? encryptionLevel, TrustManager trustManager)
+    private static bool DiffersFromDefaultConfig(EncryptionLevel? level, TrustManager trustManager)
     {
-        if (!Neo4jUri.IsSimpleUriScheme(uri))
-        {
-            throw new ArgumentException(
-                "The encryption and trust settings cannot both be set via uri scheme and driver configuration. " +
-                $"uri scheme = {uri.Scheme}, encryption = {encryptionLevel}, trust = {trustManager}");
-        }
+        return ParseEncrypted(level) || !IsDefaultTrust(trustManager);
```
AndyHeap-NeoTech, line 67: **+ssc silently overrides explicit default chain trust**

A +ssc URI scheme now silently replaces an explicitly configured default ChainTrustManager with the accept-all InsecureTrustManager, where prior releases threw — the PR's "no security downgrade" claim does not hold on the trust axis.

`GraphDatabase.Driver("neo4j+ssc://host", auth, o => o.WithTrustManager(TrustManager.CreateChainTrust()))`: `DiffersFromDefaultConfig` returns `ParseEncrypted(level) || !IsDefaultTrust(trustManager)` and `IsDefaultTrust` (line 70) treats a default-configured ChainTrustManager as unset, so control falls to `CreateFromUriScheme`, where Neo4jUri.cs:80-82 maps +ssc to `CreateInsecureTrustManager` — the user's certificate-verifying trust manager is fully discarded and every certificate is accepted. 

Pre-PR this exact combination threw ArgumentException. The mechanism is deliberate (tested by ShouldNotErrorAndUseSchemeIfTrustManagerMatchesTheDefaultChainTrust and matching java's isCustomized semantics), but the PR body's "There is no security downgrade (it can only resolve to the scheme's own encrypted setting)" addresses only the encrypted flag. 

Recommend acknowledging the trust-axis consequence explicitly in the PR/changelog — this is the strongest argument for the warning log the PR chose to defer.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Expectations/OutboundRoundTrip.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public OutboundRoundTrip(IExpectationStore expectationStore, IResponseWriter writer)
+    {
+        _expectationStore = expectationStore;
+        _writer = writer;
+    }
+
+    public async Task<T> SendExpectingAsync<T>(IProtocolMessage message, string key)
+    {
+        var value = _expectationStore.Expect<T>(key);
```
AndyHeap-NeoTech: **Expectation leaks when WriteAsync throws, poisoning key**

SendExpectingAsync registers the expectation before writing, with no cleanup on write failure, so a throwing WriteAsync permanently poisons that key for the rest of the connection.

`var expectation = _expectationStore.Expect(key); await _writer.WriteAsync(request);` — no try/catch/finally in the class, no unregister API on ExpectationStore besides Fulfil/Fail/CancelAll, and no caller cleans up. 

RetryableTransaction keys on the session id (RetryableTransaction.cs:112-116): after one failed write the driver's retry logic re-enters `RunAttemptAsync`, and `Expect(sessionId)` throws TestKitProtocolException "An expectation is already pending" (ExpectationStore.cs:43-46) for every subsequent attempt on that connection; the orphaned task's cancellation is also unobserved. 

Additionally, when the store is already cancelled, `Expect` returns Task.FromCanceled without throwing (ExpectationStore.cs:40), so the outbound request is still written to a closing connection before the await faults. 

Fix: wrap the write so a failure removes/fails the pending entry.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        DateTimeProvider.StaticInstance = _original!;
+        _original = null;
+        _fake = null;
+    }
+
+    private class FakeDateTimeProvider : IDateTimeProvider
+    {
+        private readonly List<FakeTimer> _timers = [];
```
AndyHeap-NeoTech, line 75: **Unsynchronized timer list raced by driver threads**

FakeDateTimeProvider._timers is an unsynchronized List<FakeTimer> mutated by driver threads via NewTimer() while the FakeTimeTick handler enumerates it in Advance().

`NewTimer()` does an unlocked `_timers.Add` and is called from PooledConnection constructors on driver I/O threads (PooledConnection.cs:37,39 use `DateTimeProvider.StaticInstance.NewTimer()`), while `Advance()` runs an unlocked `foreach` on a thread-pool task — MessageLoop dispatches every handler fire-and-forget, so this concurrency is the normal fake-time usage (tick while a driver call is in flight). 

Result: intermittent InvalidOperationException "Collection was modified" surfacing as a spurious BackendError, or a lost Add so the new connection's timer never advances and the timeout under test never fires — a CI-only flake. Timers are also never removed for the install's lifetime. 

Fix: lock around Add/enumerate or use a concurrent collection.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            _now = _now.AddMilliseconds(milliseconds);
+            foreach (var timer in _timers)
+            {
+                timer.Advance(milliseconds);
+            }
+        }
+    }
+
+    private class FakeTimer : ITimer
```
AndyHeap-NeoTech, line 117: **FakeTimer ignores ITimer stop semantics on Reset/Start**

FakeTimer violates the ITimer contract — Reset() zeroes without stopping, Start() is a no-op, and Advance() accrues unconditionally — so checked-out connections accumulate phantom idle time under fake time.

The contract says Reset() "Stops time interval measurement and resets the elapsed time to zero" (IPooledConnection.cs:36-38), and StopwatchBasedTimer stops accordingly. ConnectionValidator relies on that: on acquire it calls `IdleTimer.Reset()` (ConnectionValidator.cs:122-128) precisely so no idle time accrues while a connection is in use, and restarts the timer only on release. FakeTimer accrues on every Advance regardless, so the first testkit test that ticks fake time mid-transaction with a pool idle/liveness config sees in-use connections wrongly trip HasBeenIdleForTooLong or force liveness probes on release — a cross-driver parity failure that will be debugged as a pool bug rather than a shim bug. 

Fix: track running state; Reset stops, Start starts, Advance only accrues while running.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
```diff
@@ -0,0 +1,115 @@
...
+        DateTimeProvider.StaticInstance = _fake;
+    }
+
+    public void Tick(long incrementMs)
+    {
+        _fake!.Advance(incrementMs);
+    }
+
+    public void Uninstall()
+    {
```
AndyHeap-NeoTech, line 58: **Plausible? - Overlapping connections corrupt global fake clock**

Per-connection FakeTimeService swaps the process-global DateTimeProvider.StaticInstance with no ownership check, so overlapping connections corrupt each other's clock and can leave a dead fake installed for all later tests.

Install() guards only against double-install on the same instance, and Uninstall() restores `_original` unconditionally without checking `DateTimeProvider.StaticInstance == _fake`. Connection N's scope disposal runs only after RunAsync returns (after socket close and WhenAll), which can be after testkit has already opened connection N+1: if N+1 installs first, N+1 captures N's fake as `_original`; N's disposal then restores the real clock mid-test for N+1 (ticks silently stop affecting the driver), and N+1's uninstall later re-installs N's dead fake for every subsequent test. Scope-disposal cleanup (Dispose => Uninstall, pinned by FakeTimeLifetimeTests) narrows but does not close the window. 

Fix: only restore when the static instance is still this service's fake, or hold a static owner token.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Connection/MessageLoop.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            _logger.LogError(exception, "Connection {ConnectionId} failed", connectionId);
+            await _responseWriter.WriteAsync(new BackendErrorResponse { Msg = exception.Message });
+        }
+        finally
+        {
+            _expectationStore.CancelAll();
+            try
+            {
+                await Task.WhenAll(handlerTasks);
```
AndyHeap-NeoTech: **Shutdown waits forever on handler stuck in driver call**

Shutdown awaits Task.WhenAll(handlerTasks) with no timeout after CancelAll(), which only unparks expectation-awaiting handlers — a handler stuck inside a driver call hangs the connection's cleanup forever.

`_expectationStore.CancelAll(); await Task.WhenAll(handlerTasks);` — DispatchAsync carries no CancellationToken and TestkitConnectionHandler has no timeout. A handler blocked in a driver call that never completes (e.g. a retry loop under fake time: FakeTimer only advances on FakeTimeTick messages, which stop arriving once testkit closes the socket, so maxRetryTime never elapses) means RunAsync never returns, so the DI scope that would dispose the driver — the very thing that could abort the stuck call — never unwinds: a circular wait. The connection leaks, and if fake time was installed it stays installed process-globally (see the FakeTimeService findings). 

Consider a bounded WhenAll (timeout + log + proceed to scope disposal) and/or flowing a cancellation token into handlers.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/RoutingTableManager.cs
```diff
@@ -87,7 +115,12 @@ public async Task<IRoutingTable> EnsureRoutingTableForModeAsync(
         await semaphore.WaitAsync().ConfigureAwait(false);
         try
         {
-            if (_routingTables.TryGetValue(database, out var existingTable) &&
+            // discard under the lock, or a concurrently cached table can be returned unrefreshed
+            if (forceUpdate)
+            {
+                _routingTables.TryRemove(database, out _);
```
AndyHeap-NeoTech, line 121: **Forced refresh discards routing table before rediscovery**

ForceUpdateAsync removes the cached routing table before rediscovery runs, so the refresh can only bootstrap from the initial URI and a failed refresh permanently discards a still-valid table (testkit-only path).

`_routingTables.TryRemove(database, out _)` runs before `UpdateRoutingTableAsync`; RoutingTableFor(database) then returns null, an empty table is created, and rediscovery's first pass iterates zero known routers, falling back to `_initialServerAddressProvider.Get()` only (lines ~301-322). If the initial router is down but previously learned routers are alive, the forced update throws ServiceUnavailableException even though a non-destructive refresh would have succeeded — and the removed table is never restored (the finally only releases the semaphore). 

Related on the same path: the new table is accepted with only `!IsStale(AccessMode.Read)` (line 94), so a writer-less table can replace one that had a writer, and BoltProtocolAdapter.cs:79 hardcodes `sessionConfig: null`, ignoring impersonation/home-db context. Only reachable via the testkit backend (Driver.cs:193 "Used for testing with testkit only"), so no production exposure — but mark-stale-then-refresh (as other drivers do) would be more faithful for the routing tests this exists to serve.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Properties/AssemblyAttributes.cs
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Neo4j.Driver.TestKitBackend.Tests")]
```
AndyHeap-NeoTech, line 22: **Keyless InternalsVisibleTo fails signed Release build**

InternalsVisibleTo attributes without a PublicKey break the Release build (CS1726) because common.props strong-name signs every project in Release.

Reproduced with SDK 10.0.101: `dotnet build Neo4j.Driver.TestKitBackend.csproj -c Release` fails with `error CS1726: Friend assembly reference 'Neo4j.Driver.TestKitBackend.Tests' is invalid. Strong-name signed assemblies must specify a public key...` (and the same for DynamicProxyGenAssembly2), because common.props:15-19 sets SignAssembly=true for '$(Configuration)'=='Release' with no project filter and this csproj imports it without opting out. Mitigating: CI builds only the CI configuration (testkit/build.py), and the renamed legacy project carries the identical latent defect, so nothing currently green regresses — but the first Release solution build with the real neo4j.snk fails. 

Fix: adopt the `#if STRONGNAMED` keyed-IVT pattern from Neo4j.Driver/Properties/AssemblyInfo.cs:33-55, or set SignAssembly=false for this non-shipped project.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Serialization/JsonOptionsProvider.cs
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Text.Json.Serialization.Metadata;
+using Neo4j.Driver.TestKitBackend.ObjectStorage;
+
+namespace Neo4j.Driver.TestKitBackend.Serialization;
+
+[RegistrationLifetime(RegistrationLifetime.PerLifetimeScope)]
```
AndyHeap-NeoTech: **JsonSerializerOptions rebuilt per testkit connection**

JsonSerializerOptions is registered PerLifetimeScope with one scope per testkit connection, discarding System.Text.Json's per-options type-metadata cache for every one of the suite's ~6,400 test connections.

STJ caches reflection-generated JsonTypeInfo per options instance; rebuilding per connection re-runs the resolver modifiers (attribute reflection plus `Activator.CreateInstance(MakeGenericType(...))` per envelope/stored-object property) for every test. The only scope-bound dependency is IObjectStore via StoredObjectConverter — the project already solves exactly this shape with a singleton accessor publishing per-connection state (LoggingContextAccessor.cs:18-31), so a singleton options provider plus an object-store accessor is a drop-in refactor. Same theme: StoredObjectFieldTransformer's `_handleFields` reflection cache (line 32) is per-instance but the class defaults to PerDependency, so that cache never survives a connection either — mark it Singleton or make the cache static.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Messages/NewSessionConfigMapper.cs
```diff
@@ -0,0 +1,124 @@
...
+    private void ApplyNotifications(NewSessionRequest request, ISessionConfigBuilder builder)
+    {
+        _notificationsMapper.Apply(
+            request.NotificationsMinSeverity,
+            request.NotificationsDisabledCategories,
+            () => builder.WithNotificationsDisabled(),
+            (severity, categories) => builder.WithNotifications(severity, categories));
+    }
+
+    private static void ApplyRemainingProperties(NewSessionRequest request, ISessionConfigBuilder builder)
```
AndyHeap-NeoTech: **Reflection auto-mapper duplicated across config mappers**

The ApplyRemainingProperties reflection auto-mapper and its hand-maintained HandledExplicitly HashSet are duplicated as near-copies in NewSessionConfigMapper and NewDriverConfigMapper.

Two independent copies of the same ~35-line routine (property scan, `"With" + Name` lookup, TargetInvocationException unwrap) plus two HandledExplicitly sets (NewSessionConfigMapper.cs:28,94-123; NewDriverConfigMapper.cs:31,176-208, which adds an Ms-suffix-to-TimeSpan twist). Adding a request property requires remembering to update the right HashSet, and a mismatch surfaces only at runtime as a reflection failure rather than a compile error — twice, in two files that will drift. 

Either extract one shared helper (parameterizing the Ms conversion) or replace with explicit `if (request.X is { } x) builder.WithX(x);` one-liners in the style of the surrounding Apply* methods, which is roughly line-count-neutral.

Neo4j.Driver/Neo4j.Driver/Internal/EncryptionManager.cs
RichardIrons-neo4j: Agreed on all counts. The PR body's "no security downgrade" sentence was wrong on the trust axis and has been rewritten: it now states that an explicitly configured default `ChainTrustManager` combined with `+ssc` silently resolves to the accept-all `InsecureTrustManager` where prior releases threw, and names that as the strongest argument for a warning log on the silent-override path. The behaviour itself stays as testkit's contract requires. The warning log remains deferred, recorded in the body as a known, accepted gap.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Expectations/OutboundRoundTrip.cs
RichardIrons-neo4j: Fixed in a8376357. `SendExpectingAsync` now returns the cancelled task straight away when the store is already cancelled, so nothing is written to a closing connection. A write failure calls `Fail(key, exception)` on the pending entry, which frees the key and surfaces the writer's exception through the await instead of the later "expectation is already pending" you traced. Pinned by two tests in `OutboundRoundTripTests`: a failed write rethrows the writer's exception and leaves the key reusable, and a cancelled store never reaches the writer.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
RichardIrons-neo4j: Fixed in a8376357. `NewTimer`, `Advance`, and `Now()` all take the same `Lock` now. A stress test (`Timers_can_be_created_while_a_tick_is_advancing`, 2000 ticks racing 2000 timer creations) reproduced the `InvalidOperationException` before the fix and is green after. On the never-removed point: the fake provider is created per install and the install is per test, so the list is bounded by one test's connections.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
RichardIrons-neo4j: Fixed in a8376357. `FakeTimer` now mirrors `StopwatchBasedTimer`: created stopped, `Start()` runs it, `Reset()` stops and zeroes, and `Advance` only accrues while running. Worth noting the shipped behaviour was exact legacy parity (legacy's `FakeTimer` has the same no-op `Start` and unconditional accrue), so this is a deliberate divergence in favour of the `ITimer` contract rather than a regression fix. Pinned by three timer tests in `FakeTimeServiceTests`; the driver_parameters (57) and authorization (168) stub modules pass against the new semantics, which covers the liveness and token-expiry paths that consume these timers.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Time/FakeTimeService.cs
RichardIrons-neo4j: Fixed in a8376357, and "plausible" is right: the window is real but needs test N's teardown (which can trail its socket close by up to the handler drain) to overlap test N+1's install. Each fake now carries the provider it replaced, skipping over a superseded fake when capturing it, and `Uninstall` restores only while its own fake is still installed. The A-install/B-install/A-uninstall/B-uninstall sequence now ends at the real clock, so a dead fake can no longer stay installed for the rest of the suite. Pinned by the two overlap tests in `FakeTimeServiceTests`.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Connection/MessageLoop.cs
RichardIrons-neo4j: Fixed in a8376357. The drain is now `Task.WhenAll(handlerTasks).WaitAsync(HandlerDrainTimeout)` (10s default); on timeout the loop logs a warning and proceeds, so scope disposal always runs. The circular wait you describe was real: disposal was the only thing that could unstick the handler, and the handler was blocking disposal. A cancellation token into handlers would be the fuller fix, but the blocking points are inside driver calls that don't take one, so the bounded drain is where the value is.

Neo4j.Driver/Neo4j.Driver/Internal/Routing/RoutingTableManager.cs
RichardIrons-neo4j: Deferred to DRIVERS-523 rather than changed in this PR. All routing stub tests pass against the discard-first implementation, and as you note the path is testkit-only, so swapping the rediscovery semantics this late felt like regression risk for fidelity no test currently demands. The ticket carries mark-stale-then-refresh plus both sub-points (the writer-less table acceptance and the hardcoded `sessionConfig: null`).

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Properties/AssemblyAttributes.cs
RichardIrons-neo4j: Fixed in a8376357 with the `#if STRONGNAMED` keyed pattern from the driver's own `AssemblyInfo`, plus the STRONGNAMED define on the Release configuration in the csproj. `SignAssembly=false` was the tempting one-liner but it's wrong here: the driver's STRONGNAMED IVT list already grants to a keyed `Neo4j.Driver.TestKitBackend`, so Release expects the backend signed, and unsigning it would break the backend's driver-internals access instead. Verified as far as possible without the private snk: CI builds clean, and force-defining STRONGNAMED without signing fails with CS0281 from the driver's keyed grant, which is the Release signing contract engaging as it should. Legacy's identical latent defect is left as-is.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Serialization/JsonOptionsProvider.cs
RichardIrons-neo4j: Fixed across 1ffd8977 and dacacbd5. The first pass matched the shape you suggested: singleton options plus an AsyncLocal accessor mirroring `LoggingContextAccessor`, with `StoredObjectFieldTransformer` made a singleton so its reflection cache survives. Then dacacbd5 went a step further: the object store itself is now a process singleton that the message loop clears at connection end (placed before the handler drain, so a stuck handler can't delay the clear into the next test), which dissolved the accessor entirely and lets the converters take `IObjectStore` directly. Options are built once per process, pinned by a container test asserting `GetOptions()` is reference-equal across connection scopes.

Neo4j.Driver/Neo4j.Driver.TestKitBackend/Messages/NewSessionConfigMapper.cs
RichardIrons-neo4j: Fixed in 0e2c3cd7. Went with the shared helper (`IRemainingPropertiesMapper`) rather than explicit `WithX` one-liners, mainly to keep the fail-loud property: a request property with no matching builder method still throws naming both, so a mismatch surfaces the first time testkit sends the field instead of being silently dropped. The Ms-to-TimeSpan rule ended up unconditional rather than parameterised, since the session request has no Ms-suffixed auto-mapped properties. Each mapper keeps its own `HandledExplicitly` set.

AndyHeap-NeoTech (approved):
Looks good.

PR #958, opened 2026-08-18
Address issues in security report

RichardIrons-neo4j:
Fixes DRIVERS-517

Response to the 6.x security audit report (30 July 2026). Four findings were raised: 3 Medium, 1 Low. Two are fixed, one is fixed in part and documented, one is deferred. Rationale for every decision is below, including the two sub-fixes we are deliberately **not** taking.

| Finding | Severity | Outcome |
|---|---|---|
| SEC-01 — custom auth token prints credentials | Medium (5.7) | Fixed |
| SEC-04 — validator accepts peer with no certificate | Medium (5.9) | Fixed |
| SEC-03 — cross-identity home-database misrouting | Medium (5.3) | 1 of 3 sub-fixes taken; 2 rejected; contract documented |
| SEC-05 — query payloads in debug log | Low (2.3) | Deferred |

Every finding was first re-verified against `6.x` at `b8091bee`. The report notes:

> **Version control was unavailable.** The directory reviewed is not a git checkout, so we could not check whether any finding has already been fixed on a branch

Nothing had been fixed in the interim — every line reference and code quote in the report matched the branch exactly.

The report also asks:

> **Static review only.** Nothing was executed or exploited. Reproduction steps are written procedures we have reasoned through, not runs we performed. We recommend confirming each one before scheduling remediation.

Each finding below was reproduced with a failing test before being fixed, and the two rejected sub-fixes were rejected on the basis of test evidence, not opinion.

---

## SEC-01 — Custom auth token prints credentials in clear text — **FIXED**

> One class — and it is the class most likely to hold a high-value secret — does the opposite.

> every other credential type in the driver deliberately hides the secret

Confirmed exactly as described: `CustomAuthToken.ToString()` excluded `scheme`/`principal`/`realm` from the "extra keys" list but not `credentials` or `parameters`, so both were printed raw.

`credentials` and `parameters` are now masked as `******`, matching the convention already used by `LogonMessage`/`HelloMessage`. Regression test asserts neither the credential nor a parameter value appears in `ToString()` output.

We accepted the report's reasoning on priority despite this being the second-ranked finding by CVSS:

> unlike anything else in this report, a leaked long-lived credential cannot be undone by deploying a patch

**Not taken:** the report's suggestion to also collapse `BearerAuthToken`/`KerberosAuthToken`'s five-character prefix to a full mask. That prefix is load-bearing for support diagnostics, and changing it is a behaviour change to two more token types that is out of scope here. Worth its own ticket if we want it.

## SEC-04 — Certificate validator accepts a peer that presented no certificate — **FIXED**

> a validator whose documented job is to verify certificates should not depend on its caller to reject a peer that presented none

Confirmed. `ChainTrustManager` tested only `RemoteCertificateChainErrors` and never asked whether a certificate was presented at all.

One correction to the report's reproduction steps, which predicted:

> **Expected today:** `trusted` is `true`. The correct result is `false`.

On macOS / .NET 8 it does not return `true` — it throws `ArgumentException: The chain context handle is invalid`, because the revocation-policy comparison sends it into `BuildChain` with a null certificate. Different symptom, same defect: it fails to cleanly refuse. The other three validators throw `NullReferenceException`. All four were wrong, none deliberately.

Implemented as the report recommended:

> Make every validator self-sufficient rather than relying on its caller. Add the same two-line precondition to all four implementations of `ValidateServerCertificate`.

The precondition lives in one `internal` helper on the `TrustManager` base class rather than being copy-pasted four times, so validators added later inherit it, and nothing is added to the public API surface. This is behaviour-neutral for every built-in path, because `DefaultTlsNegotiator` already rejects `RemoteCertificateNotAvailable` before any validator runs — it only closes the gap for applications supplying a custom `ITlsNegotiator`.

Because the helper is `internal`, it does not reach a third-party `TrustManager`, and `TrustManager` is public and abstract, so subclassing it is supported. The precondition is therefore stated on `ValidateServerCertificate` itself: an implementation owns it, and callers do not screen the case out beforehand. Widening the helper to `protected` was the alternative and was rejected to keep the public surface unchanged.

Tests cover all four validators with hostname verification on and off, for a missing certificate both with and without the policy flag, for a certificate presented alongside the flag — the one case whose result changes — and one happy path, so that a validator refusing everything no longer satisfies the file.

`InsecureTrustManager` is included, which deserves a note since `TrustManager.CreateInsecure` is documented as accepting "any certificate without validation". This is not a breaking change:

- With no certificate presented, the old code dereferenced `certificate.Subject` while building its log line and threw `NullReferenceException`. The connection failed then and fails now — a crash is replaced by a clean, logged refusal.
- The only case whose result actually flips is a non-null certificate arriving *together with* `RemoteCertificateNotAvailable`, which is self-contradictory and not reachable in practice.

"Insecure" means the driver will not judge which certificate to trust; it does not mean authenticating a peer that presented nothing at all.

## SEC-03 — Cross-identity database misrouting via the home-database cache — **PARTIALLY FIXED**

The report was explicit that this finding hinged on an unresolved question:

> Whether this is a live defect or defence-in-depth hardening turns on one unanswered question, and we could not resolve it from the code.

> The scenario assumes an application uses `IAuthTokenManager` to rotate between **different users**. If it is instead used to refresh **one user's** expiring token — which is what the built-in `Neo4jAuthTokenManager` does — the current behaviour is correct and **there is no bug**.

**That question is now settled, and the answer is that identity rotation is not supported.** Testkit specifies the opposite of the report's assumption, in `tests/stub/homedb/test_homedb.py::test_cache_considers_all_driver_auth_equal`:

> ```
> # Then: regardless of the token returned by the auth
> #       manager the driver should use the cached home db
> ```

That test rotates a driver-level `IAuthTokenManager` across three distinct users and asserts the cached home database is reused with **no new ROUTE request**. Driver-level auth is deliberately excluded from the cache key across all drivers; per-identity partitioning is expected to come from session-level auth or impersonation, both of which are already keyed correctly today.

So the report's own alternate branch applies:

> **If it is not supported** — the three fixes below are still correct hardening, but the urgency drops considerably, and the interface should be documented to say so explicitly.

We disagree that (a) and (c) are "still correct hardening" — measured against testkit they are regressions, not hardening. Each sub-fix was implemented and measured independently against `tests/stub/homedb` (baseline: 37 passed):

| Sub-fix | Result | Decision |
|---|---|---|
| (a) put identity in the cache key | 3 failures | **Rejected** |
| (b) count live connections, not borrow events | 37 passed | **Fixed** |
| (c) honour `isCachedHomeDb` on the early-return path | 39 failures, 4 errors | **Rejected** |

### (a) Rejected — contradicts the cross-driver contract

Fails `test_cache_considers_all_driver_auth_equal` in all three session variants. Partitioning the cache by driver-level token is precisely what testkit forbids.

It also required fetching a token where none was previously needed, which broke `tests/stub/authorization/test_auth_token_manager` — 160 failures, all routed cases, `get_auth_count` 3 where 2 was expected. Making the fetch lazy fixed that symptom, but the underlying spec conflict remained, so the sub-fix is dropped entirely.

### (b) Fixed — a real counter bug, independent of the identity question

> The safety gate that should disable the cache reads a meaningless number.

> It is therefore "borrows minus destructions", which bears no relationship to how many connections exist.

Confirmed and reproduced: destroying a connection that never completed `AddConnectionAsync` (a routine failed connection attempt) drives `NumberOfConnectionsWithSsrDisabled` to `-1`, and both consumers compare against exact constants, so drift can produce a false zero and re-enable the home-db cache in a mixed cluster that should have it disabled.

The report also observed that the codebase already contained a correct answer to the same question:

> **Corroborating evidence that this is a defect, not a design:** the codebase already contains a *correct* implementation of the same question — `ConnectionPool.IsOnlyConnectionWithoutSsr` — which counts the live connection collections directly. Nothing calls it.

That method is now deleted rather than promoted. The question it answers is cluster-wide, which a single pool cannot see, and `ClusterConnectionPool.ConnectionCausesCacheDisable` already answers it at that level; keeping a second per-pool implementation only invited a fix in the wrong place.

The fix is that a connection is now tracked from the end of a successful init until destruction — the same lifetime as `_poolSize` — and membership decides whether a count is owed, so the pair cannot drift the way the borrow/destroy pair did. A unit test asserts the count stays at `0` when a connection fails before entering the pool.

Deriving the counts from the `_inUseConnections`/`_idleConnections` collections, which is what this PR did first, is **not** a correct substitute, and review caught it: `ReleaseAsync` removes from in-use, awaits a RESET on the wire, then adds to idle, so for the duration of a server round trip a live connection belongs to neither collection. `CanUseHomeDbCache` would then read zero SSR-disabled connections and enable the home-db cache in a mixed cluster — failing unsafe, where the counters it replaced had only ever over-reported. The acquire side has the same gap around the liveness probe. Tracking membership over the whole lifetime avoids both windows, and keeps these reads at two `int` loads rather than a locked snapshot of every hash bucket on each connection acquisition.

### (c) Rejected — defeats the purpose of the cache it is protecting

> The re-check that should ask the server again is skipped.

The behaviour is real, but forcing a fresh discovery whenever the database name came from the home-db cache makes every cached acquisition perform the exact round trip the cache exists to avoid. It fails `assert_no_new_route_request` across the board (`3 != 2` ROUTE requests) and breaks `TestHomeDbMixedCluster` home-db fallback.

With (a) rejected, the premise for (c) is also gone: the cached name is per-spec correct for the identities the cache is keyed on, so a routing table keyed by that name is the right routing table. The report offered these as independent:

> Any *one* of these breaks the chain. We recommend all three

Taking (b) alone is sufficient for the defect that actually exists.

### Tests pinning the specified behaviour

Rather than delete the tests written while exploring (a) and (c), they are inverted to lock in the contract testkit specifies, so a future change in this area fails fast at unit level:

- `HomeDbCacheKeyProviderTests` — all driver-level auth tokens share one cache entry; session-level auth and impersonation each partition; a session carrying both keys on the impersonated user
- `RoutingTableManagerTests` — a fresh routing table reached through a cached home database is reused with no discovery, and a stale one is rediscovered with the cached name withheld, which is the only thing `isCachedHomeDb` does

### Documenting the contract

The report's remaining recommendation for this branch of the decision:

> the interface should be documented to say so explicitly

`IAuthTokenManager` now carries a `<remarks>` block stating that an implementation supplies credentials for a single identity and renews them, that it must not be used to switch users, and that the driver caches per-identity state such as the resolved home database without regard to the token it returns. It points callers who need multiple identities at `WithAuthToken` or `WithImpersonatedUser`, which are keyed correctly.

A `<remarks>` on the interface is not where the mistake gets made, though, so the constraint is cross-referenced from the two places an application author actually chooses a manager: the `AuthTokenManagers` factories and the `GraphDatabase.Driver` overloads that take one.

This is the part of SEC-03 that closes the actual exposure. The behaviour is specified and tested; what was missing was any statement of it on the public interface, which is why the report could not resolve the question from the source and why an application author could reasonably have assumed rotation was supported.

## SEC-05 — Query text, parameters and results written to the debug log — **DEFERRED**

Confirmed as described. Not addressed here, and the report itself is ambivalent:

> **This is the weakest finding in the report, and it is arguably not a defect at all.** Verbose payload logging at debug level is conventional across database drivers, and credential fields specifically *are* redacted.

> If you read it as squarely within the configuration-choice exclusion, drop it and the report stands at three findings.

The substantive part is a documentation gap rather than a code defect, and the proposed `LogQueryParameters` opt-out defaulting to redacted is a behaviour change to debug output that deserves its own discussion. Being handled separately.

---

## Testing

Unit tests, per target framework (net8.0 / net9.0 / net10.0):

- 2757 passed, 0 failed, 6 skipped (pre-existing platform-specific skips)

Testkit stub tests (testkit `6.x` at `e06789d`, Python stub server):

- `tests/stub/authorization/test_auth_token_manager` — 16 passed
- `tests/stub/homedb` — 37 passed, 7 skipped (identical to the pre-change baseline)
- `tests/stub/server_side_routing` — 2 passed

Three of the test changes here fix coverage rather than behaviour, so passing proves nothing on its own. Each was checked by mutation instead — the trust-manager happy path fails if the missing-certificate guard is made to refuse unconditionally, the routing test fails if `isCachedHomeDb` is ignored, and the cache-key precedence test fails if impersonation and session-level auth are swapped.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Security;
+using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using FluentAssertions;
 using Neo4j.Driver.Internal.Logging;
@@ -38,6 +39,13 @@ public class TrustManagerTests
         }
     }
 
+    private static X509Certificate2 SelfSignedCertificate()
+    {
+        using var key = RSA.Create(2048);
+        var request = new CertificateRequest("CN=example.test", key, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+        return request.CreateSelfSigned(DateTimeOffset.UtcNow.AddDays(-1), DateTimeOffset.UtcNow.AddDays(1));
+    }
+
     [Theory]
     [MemberData(nameof(AllTrustManagers))]
     public void ShouldRefusePeerThatPresentedNoCertificate(TrustManager trustManager, bool verifyHostname)
@@ -51,5 +59,52 @@ public class TrustManagerTests
             SslPolicyErrors.RemoteCertificateNotAvailable);
 
         trusted.Should().BeFalse($"verifyHostname: {verifyHostname}");
+    }
+
+    [Theory]
+    [MemberData(nameof(AllTrustManagers))]
+    public void ShouldRefuseCertificateAccompaniedByNotAvailableFlag(TrustManager trustManager, bool verifyHostname)
+    {
+        trustManager.Neo4JLogger = NullNeo4JLogger.Instance;
+        using var certificate = SelfSignedCertificate();
+
+        var trusted = trustManager.ValidateServerCertificate(
+            ServerUri,
+            certificate,
+            new X509Chain(),
+            SslPolicyErrors.RemoteCertificateNotAvailable);
+
+        trusted.Should().BeFalse($"verifyHostname: {verifyHostname}");
+    }
+
+    [Theory]
+    [MemberData(nameof(AllTrustManagers))]
+    public void ShouldRefuseMissingCertificateEvenWithoutPolicyErrors(TrustManager trustManager, bool verifyHostname)
+    {
+        trustManager.Neo4JLogger = NullNeo4JLogger.Instance;
+
+        var trusted = trustManager.ValidateServerCertificate(
+            ServerUri,
+            null,
+            new X509Chain(),
+            SslPolicyErrors.None);
+
+        trusted.Should().BeFalse($"verifyHostname: {verifyHostname}");
+    }
+
+    [Fact]
+    public void ShouldTrustPresentedCertificateWhenValidationIsDisabled()
+    {
+        var trustManager = TrustManager.CreateInsecure();
+        trustManager.Neo4JLogger = NullNeo4JLogger.Instance;
+        using var certificate = SelfSignedCertificate();
+
+        var trusted = trustManager.ValidateServerCertificate(
+            ServerUri,
+            certificate,
+            new X509Chain(),
+            SslPolicyErrors.None);
+
+        trusted.Should().BeTrue();
     }
 }
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) "Neo4j"
+// Neo4j Sweden AB [https://neo4j.com]
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using FluentAssertions;
+using Neo4j.Driver.Internal.Logging;
+using Xunit;
+
+namespace Neo4j.Driver.Tests.Connector.Trust;
+
+public class TrustManagerTests
+{
+    private static readonly Uri ServerUri = new("bolt://example.test:7687");
+
+    public static IEnumerable<object[]> AllTrustManagers()
+    {
+        foreach (var verifyHostname in new[] { false, true })
+        {
+            yield return [TrustManager.CreateChainTrust(verifyHostname), verifyHostname];
+            yield return [TrustManager.CreatePeerTrust(verifyHostname), verifyHostname];
+            yield return [TrustManager.CreateCertTrust([], verifyHostname), verifyHostname];
+            yield return [TrustManager.CreateInsecure(verifyHostname), verifyHostname];
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(AllTrustManagers))]
+    public void ShouldRefusePeerThatPresentedNoCertificate(TrustManager trustManager, bool verifyHostname)
+    {
+        trustManager.Neo4JLogger = NullNeo4JLogger.Instance;
+
+        var trusted = trustManager.ValidateServerCertificate(
+            ServerUri,
+            null,
+            new X509Chain(),
+            SslPolicyErrors.RemoteCertificateNotAvailable);
+
+        trusted.Should().BeFalse($"verifyHostname: {verifyHostname}");
+    }
+}
--- a/Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
@@ -58,13 +58,20 @@ public class HomeDbCacheKeyProviderTests
     }
 
     [Fact]
-    public void ShouldPreferSessionLevelAuthTokenOverDriverLevelAuthToken()
+    public void ShouldPreferImpersonatedUserOverSessionLevelAuthToken()
     {
-        var sessionConfig = SessionConfig.Builder.WithAuthToken(AuthTokens.Basic("alice", "alicepw")).Build();
+        var sessionToken = AuthTokens.Basic("bob", "bobpw");
+        var bothConfig = SessionConfig.Builder
+            .WithImpersonatedUser("alice")
+            .WithAuthToken(sessionToken)
+            .Build();
 
-        var withDriverToken = HomeDbCacheKeyProvider.GetCacheKey(AuthTokens.Basic("bob", "bobpw"), sessionConfig);
-        var withoutDriverToken = HomeDbCacheKeyProvider.GetCacheKey(null, sessionConfig);
+        var key = HomeDbCacheKeyProvider.GetCacheKey(null, bothConfig);
 
-        withDriverToken.Should().Be(withoutDriverToken);
+        var impersonationOnly = SessionConfig.Builder.WithImpersonatedUser("alice").Build();
+        var sessionTokenOnly = SessionConfig.Builder.WithAuthToken(sessionToken).Build();
+
+        key.Should().Be(HomeDbCacheKeyProvider.GetCacheKey(null, impersonationOnly));
+        key.Should().NotBe(HomeDbCacheKeyProvider.GetCacheKey(null, sessionTokenOnly));
     }
 }
--- a/Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) "Neo4j"
+// Neo4j Sweden AB [https://neo4j.com]
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using FluentAssertions;
+using Neo4j.Driver.Internal.HomeDbCaching;
+using Xunit;
+
+namespace Neo4j.Driver.Tests.HomeDbCaching;
+
+public class HomeDbCacheKeyProviderTests
+{
+    [Fact]
+    public void ShouldTreatAllDriverLevelAuthTokensAsEqual()
+    {
+        var aliceKey = HomeDbCacheKeyProvider.GetCacheKey(AuthTokens.Basic("alice", "alicepw"), null);
+        var bobKey = HomeDbCacheKeyProvider.GetCacheKey(AuthTokens.Basic("bob", "bobpw"), null);
+
+        aliceKey.Should().Be(HomeDbCacheKey.Default);
+        bobKey.Should().Be(HomeDbCacheKey.Default);
+    }
+
+    [Fact]
+    public void ShouldPartitionBySessionLevelAuthToken()
+    {
+        var aliceConfig = SessionConfig.Builder.WithAuthToken(AuthTokens.Basic("alice", "alicepw")).Build();
+        var bobConfig = SessionConfig.Builder.WithAuthToken(AuthTokens.Basic("bob", "bobpw")).Build();
+
+        var aliceKey = HomeDbCacheKeyProvider.GetCacheKey(null, aliceConfig);
+        var bobKey = HomeDbCacheKeyProvider.GetCacheKey(null, bobConfig);
+
+        aliceKey.Should().NotBe(bobKey);
+        aliceKey.Should().NotBe(HomeDbCacheKey.Default);
+    }
+
+    [Fact]
+    public void ShouldPartitionByImpersonatedUser()
+    {
+        var aliceConfig = SessionConfig.Builder.WithImpersonatedUser("alice").Build();
+        var bobConfig = SessionConfig.Builder.WithImpersonatedUser("bob").Build();
+
+        var aliceKey = HomeDbCacheKeyProvider.GetCacheKey(null, aliceConfig);
+        var bobKey = HomeDbCacheKeyProvider.GetCacheKey(null, bobConfig);
+
+        aliceKey.Should().NotBe(bobKey);
+        aliceKey.Should().NotBe(HomeDbCacheKey.Default);
+    }
+
+    [Fact]
+    public void ShouldPreferSessionLevelAuthTokenOverDriverLevelAuthToken()
+    {
+        var sessionConfig = SessionConfig.Builder.WithAuthToken(AuthTokens.Basic("alice", "alicepw")).Build();
+
+        var withDriverToken = HomeDbCacheKeyProvider.GetCacheKey(AuthTokens.Basic("bob", "bobpw"), sessionConfig);
+        var withoutDriverToken = HomeDbCacheKeyProvider.GetCacheKey(null, sessionConfig);
+
+        withDriverToken.Should().Be(withoutDriverToken);
+    }
+}
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
@@ -1046,17 +1046,10 @@ public static class RoutingTableManagerTests
 
             var discovery = new Mock<IDiscovery>();
 
-            var poolManager = new Mock<IClusterConnectionPoolManager>();
-            poolManager.Setup(x => x.CreateClusterConnectionAsync(It.IsAny<Uri>(), It.IsAny<SessionConfig>()))
-                .ReturnsAsync(Mock.Of<IConnection>);
-
-            var initialAddressProvider = new Mock<IInitialServerAddressProvider>();
-            initialAddressProvider.Setup(x => x.Get()).Returns(new HashSet<Uri> { server01 });
-
             var manager = new RoutingTableManager(
-                initialAddressProvider.Object,
+                Mock.Of<IInitialServerAddressProvider>(),
                 discovery.Object,
-                poolManager.Object,
+                Mock.Of<IClusterConnectionPoolManager>(),
                 Mock.Of<INeo4jLogger>(),
                 TimeSpan.MaxValue,
                 cachedHomeDbTable);
@@ -1073,6 +1066,68 @@ public static class RoutingTableManagerTests
                 x => x.DiscoverAsync(
                     It.IsAny<IConnection>(),
                     It.IsAny<string>(),
+                    It.IsAny<SessionConfig>(),
+                    It.IsAny<Bookmarks>(),
+                    It.IsAny<IHomeDbCache>()),
+                Times.Never);
+        }
+
+        [Fact]
+        public async Task ShouldRediscoverHomeDatabaseWhenCachedNameIsStale()
+        {
+            var staleTable = new Mock<IRoutingTable>();
+            staleTable.Setup(x => x.Database).Returns("db_alice");
+            staleTable.Setup(x => x.IsStale(It.IsAny<AccessMode>())).Returns(true);
+            staleTable.Setup(x => x.Routers).Returns(new List<Uri> { server01 });
+
+            var freshTable =
+                new RoutingTable("db_bob", new[] { server04 }, new[] { server05 }, new[] { server06 }, 1000);
+
+            var discovery = new Mock<IDiscovery>();
+            discovery.Setup(
+                    x => x.DiscoverAsync(
+                        It.IsAny<IConnection>(),
+                        It.IsAny<string>(),
+                        It.IsAny<SessionConfig>(),
+                        It.IsAny<Bookmarks>(),
+                        It.IsAny<IHomeDbCache>()))
+                .ReturnsAsync(freshTable);
+
+            var poolManager = new Mock<IClusterConnectionPoolManager>();
+            poolManager.Setup(x => x.CreateClusterConnectionAsync(It.IsAny<Uri>(), It.IsAny<SessionConfig>()))
+                .ReturnsAsync(Mock.Of<IConnection>);
+
+            var initialAddressProvider = new Mock<IInitialServerAddressProvider>();
+            initialAddressProvider.Setup(x => x.Get()).Returns(new HashSet<Uri> { server01 });
+
+            var manager = new RoutingTableManager(
+                initialAddressProvider.Object,
+                discovery.Object,
+                poolManager.Object,
+                Mock.Of<INeo4jLogger>(),
+                TimeSpan.MaxValue,
+                staleTable.Object);
+
+            await manager.EnsureRoutingTableForModeAsync(
+                AccessMode.Read,
+                "db_alice",
+                true,
+                null,
+                Bookmarks.Empty);
+
+            discovery.Verify(
+                x => x.DiscoverAsync(
+                    It.IsAny<IConnection>(),
+                    "",
+                    It.IsAny<SessionConfig>(),
+                    It.IsAny<Bookmarks>(),
+                    It.IsAny<IHomeDbCache>()),
+                Times.Once);
+
+            discovery.Verify(
+                x => x.DiscoverAsync(
+                    It.IsAny<IConnection>(),
+                    "db_alice",
                     It.IsAny<SessionConfig>(),
                     It.IsAny<Bookmarks>(),
                     It.IsAny<IHomeDbCache>()),
--- a/Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
+++ b/Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
@@ -1039,6 +1039,47 @@ public static class RoutingTableManagerTests
         }
 
         [Fact]
+        public async Task ShouldReuseExistingRoutingTableWhenDatabaseCameFromHomeDbCache()
+        {
+            var cachedHomeDbTable =
+                new RoutingTable("db_alice", new[] { server01 }, new[] { server02 }, new[] { server03 }, 1000);
+
+            var discovery = new Mock<IDiscovery>();
+
+            var poolManager = new Mock<IClusterConnectionPoolManager>();
+            poolManager.Setup(x => x.CreateClusterConnectionAsync(It.IsAny<Uri>(), It.IsAny<SessionConfig>()))
+                .ReturnsAsync(Mock.Of<IConnection>);
+
+            var initialAddressProvider = new Mock<IInitialServerAddressProvider>();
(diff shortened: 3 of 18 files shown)


Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -116,9 +113,9 @@ internal ConnectionPool(
     internal int PoolSize => Interlocked.CompareExchange(ref _poolSize, -1, -1);
     public int NumberOfInUseConnections => _inUseConnections.Count;
     public int NumberOfIdleConnections => _idleConnections.Count;
-    public int NumberOfConnectionsWithSsrEnabled => _connectionsWithSsrEnabled;
-    public int NumberOfConnectionsWithSsrDisabled => _connectionsWithSsrDisabled;
-    public int TotalNumberOfConnections => _connectionsWithSsrDisabled + _connectionsWithSsrEnabled;
+    public int NumberOfConnectionsWithSsrEnabled => _inUseConnections.Concat(_idleConnections).Count(c => c.SsrEnabled);
```
AndyHeap-NeoTech: **1. Derived SSR counts under-report while a connection is mid-release, which can re-enable the home-db cache in a mixed cluster**

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs:116-118

```
public int NumberOfConnectionsWithSsrEnabled => _inUseConnections.Concat(_idleConnections).Count(c => c.SsrEnabled);
public int NumberOfConnectionsWithSsrDisabled => _inUseConnections.Concat(_idleConnections).Count(c => !c.SsrEnabled);
public int TotalNumberOfConnections => _inUseConnections.Count + _idleConnections.Count;
```

The population "in-use ∪ idle" is not the population "live". ReleaseAsync removes from one collection and adds to the other with a network round trip in between (ConnectionPool.cs:197-216):

```
if (!_inUseConnections.TryRemove(connection))     // <-- 197: leaves in-use
{
    return;
}

if (!await _connectionValidator.OnReleaseAsync(connection).ConfigureAwait(false) ||   // <-- RESET on the wire
    IsIdlePoolFull() ||
    IsInactiveOrClosed)
{
    await DestroyConnectionAsync(connection).ConfigureAwait(false);
    return;
}
...
_idleConnections.Add(connection);                 // <-- 216: rejoins idle
```

The same gap exists on the acquire side: GetPooledOrNewConnectionAsync does _idleConnections.TryTake(...) (:547) and the connection only re-enters via AddConnectionAsync → _inUseConnections.TryAdd (:526) after the liveness probe, which is another ResetAsync/SyncAsync round trip (:508-513).

So for the duration of a server round trip, a live connection is invisible. Now look at the consumer that must not under-count — Internal/Routing/ClusterConnectionPool.cs:149-160:

```
public bool CanUseHomeDbCache()
{
    var ssrEnabledCount = 0;
    var ssrDisabledCount = 0;
    foreach (var pool in _pools.Values)
    {
        ssrEnabledCount += pool.NumberOfConnectionsWithSsrEnabled;
        ssrDisabledCount += pool.NumberOfConnectionsWithSsrDisabled;
    }

    return ssrEnabledCount >= 1 && ssrDisabledCount == 0;
}
```

Failure scenario: mixed cluster, one SSR-disabled connection and one or more SSR-enabled. Thread A releases the SSR-disabled connection and is awaiting its RESET. Thread B enters LoadBalancer.AcquireConnectionAsync (LoadBalancer.cs:292) with no database named, calls CanUseHomeDbCache(), observes ssrDisabled == 0 && ssrEnabled >= 1, returns true, and routes off the home-db cache in a cluster that should have it disabled. The pre-fix counters did not have this window — a connection was counted from first borrow until destroy (over-counted, but never under). ConnectionCausesCacheDisable fails in the safe direction, so this one path is the exposure.

TotalNumberOfConnections inherits the same transient under-report; ClusterConnectionPool.cs:135-138 sums it.

Suggested fix — track live connections in one collection with whole-lifetime membership, rather than inferring from the two location-specific ones. _poolSize already has exactly the right lifetime semantics (TryIncrementPoolSize / DecrementPoolSize in DestroyConnectionAsync), so the symmetric pair for SSR is: add to a ConcurrentHashSet<IPooledConnection> _liveConnections (or increment a counter) once after InitAsync succeeds, remove/decrement once in DestroyConnectionAsync when conn != null. That keeps the regression test in this PR green — a connection that fails init is never added — and removes the drift the finding was about, without introducing the release window.


**2. Same fix removes a per-acquisition cost that was previously two int reads**

Folding this in because one change addresses both. ConcurrentHashSet.GetEnumerator() (Internal/Util/ConcurrentHashSet.cs:28-31) returns _items.Keys.GetEnumerator(), and ConcurrentDictionary<,>.Keys acquires every bucket lock and copies the whole table into a new list. Each of the two SSR properties therefore takes an all-table lock and allocates a snapshot.

Per connection acquisition in routing mode, LoadBalancer calls CanUseHomeDbCache() (:292) and ConnectionCausesCacheDisable() (:156), each of which reads both properties for every pool — four all-lock snapshots per pool, per acquisition, contending directly with the TryAdd/TryRemove on the acquisition path itself. At the default MaxConnectionPoolSize of 100 over a 3-server cluster that's ~1200 element copies and 12 lock-all-buckets operations per acquisition, replacing what used to be a handful of int field reads.

Secondary: ConnectionCausesCacheDisable reads the enabled and disabled counts in two separate passes, so the pair it tests (totalSsrEnabled > 0 && totalSsrDisabled == 1) is no longer a consistent snapshot. If you keep any enumeration-based approach, count both in a single pass and expose them together.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
AndyHeap-NeoTech: IsOnlyConnectionWithoutSsr is still dead, and now there are two implementations of the question

ConnectionPool.cs:281-300. The PR body cites this method as the corroborating evidence — "Nothing calls it" — and it is still uncalled outside tests, while ClusterConnectionPool.ConnectionCausesCacheDisable answers a near-identical question cluster-wide. Either route the cluster-level check through it or delete it along with the IConnectionPool.IsOnlyConnectionWithoutSsr member (Internal/IConnectionPool.cs:33). Leaving both invites the next reader to fix the wrong one.

Neo4j.Driver/Neo4j.Driver/Public/TrustManager.cs
```diff
@@ -41,6 +41,17 @@ public abstract class TrustManager
 {
     internal INeo4jLogger Neo4JLogger { get; set; }
 
+    internal bool HasNoCertificate(Uri uri, X509Certificate2 certificate, SslPolicyErrors sslPolicyErrors)
```
AndyHeap-NeoTech: HasNoCertificate is unavailable to third-party trust managers, which are a supported extension point

Public/TrustManager.cs:44-53

`internal bool HasNoCertificate(Uri uri, X509Certificate2 certificate, SslPolicyErrors sslPolicyErrors)`

TrustManager is public and abstract with a public abstract ValidateServerCertificate, and the test suite itself subclasses it from outside the trust namespace (Neo4j.Driver.Tests/Connector/EncryptionManagerTests.cs:163, CustomTrustManager : TrustManager). The PR's "future validators inherit it" holds only inside the assembly. Since the whole point of SEC-04 is that a validator must not rely on its caller, the precondition should at minimum be stated in the `<summary>/<param>` docs on ValidateServerCertificate so external implementers know they own it. protected would be the alternative, at the cost of public API surface — given the team's conservatism about surface, documenting is probably the right trade, but it should be a conscious one.

Naming, separately: HasNoCertificate reads as a pure predicate but logs an Error as a side effect. ShouldRefuseMissingCertificate or TryRejectMissingCertificate would signal that. (Unguarded Neo4JLogger deref is fine here — it matches the four existing managers and EncryptionManager.cs:76-111 guarantees assignment.)

Neo4j.Driver/Neo4j.Driver/Public/TrustManager.cs
AndyHeap-NeoTech, line 79: CreateInsecure's XML doc still promises the old behaviour

Public/TrustManager.cs:70-81 still says "accepts any certificate without validation" with no mention of the new refusal. The PR body argues carefully that narrowing this is not a breaking change — I agree with the argument — but the documented contract is the thing users rely on, and it is the one artefact that now contradicts the code. One sentence: a peer that presents no certificate is still refused.

Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            yield return [TrustManager.CreateChainTrust(verifyHostname), verifyHostname];
+            yield return [TrustManager.CreatePeerTrust(verifyHostname), verifyHostname];
+            yield return [TrustManager.CreateCertTrust([], verifyHostname), verifyHostname];
+            yield return [TrustManager.CreateInsecure(verifyHostname), verifyHostname];
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(AllTrustManagers))]
```
AndyHeap-NeoTech, line 50: TrustManagerTests misses the only case whose result actually flips

Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs:42-54 covers certificate: null + RemoteCertificateNotAvailable across all four managers. Not covered:

- non-null certificate with RemoteCertificateNotAvailable — the PR body identifies this as the only case whose result changes, so it is the one that most deserves pinning;

- null certificate without the flag — HasNoCertificate refuses this too, and that is a deliberate choice worth locking in;

- any happy path. As written, a ValidateServerCertificate that unconditionally returned false would pass this whole file.

Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
```diff
@@ -1038,6 +1038,47 @@ public void ShouldThrowOnFatalDiscovery()
                 .Be(error);
         }
 
+        [Fact]
+        public async Task ShouldReuseExistingRoutingTableWhenDatabaseCameFromHomeDbCache()
```
AndyHeap-NeoTech, line 1042: ShouldReuseExistingRoutingTableWhenDatabaseCameFromHomeDbCache does not exercise isCachedHomeDb

Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs (new test). The seeded table is fresh, so EnsureRoutingTableForModeAsync returns at the first branch:

```
if (_routingTables.TryGetValue(database, out var existingTable) &&
    !existingTable.IsStale(mode))
{
    return existingTable;          // <-- test stops here
}

var refreshedTable = await UpdateRoutingTableAsync(
        mode,
        isCachedHomeDb ? "" : database,   // <-- the parameter's only use, never reached
        ...
```

The test passes identically with isCachedHomeDb: false. It does guard against sub-fix (c) being reintroduced as an extra && !isCachedHomeDb on the early return, which is worth having — but the parameter's actual behaviour (stale table + isCachedHomeDb: true ⇒ discovery called with "") stays untested. Add that second case and the pair reads as intended. The poolManager and initialAddressProvider setups are also unreachable in this test; dropping them makes it obvious to the reader that the early path is taken.

Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        var aliceKey = HomeDbCacheKeyProvider.GetCacheKey(null, aliceConfig);
+        var bobKey = HomeDbCacheKeyProvider.GetCacheKey(null, bobConfig);
+
+        aliceKey.Should().NotBe(bobKey);
+        aliceKey.Should().NotBe(HomeDbCacheKey.Default);
+    }
+
+    [Fact]
+    public void ShouldPreferSessionLevelAuthTokenOverDriverLevelAuthToken()
```
AndyHeap-NeoTech: ShouldPreferSessionLevelAuthTokenOverDriverLevelAuthToken is tautological

Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs:61-68. HomeDbCacheKeyProvider.GetCacheKey's driverLevelAuthToken parameter is never read by the implementation, so withDriverToken == withoutDriverToken cannot fail for any reason related to precedence — there is no precedence logic involving it. ShouldTreatAllDriverLevelAuthTokensAsEqual earns its place (it fails the moment someone starts using the parameter, i.e. re-introduces sub-fix (a)); this one asserts a property of an ignored argument.

The precedence that does exist and is untested is impersonation over session-level token — GetCacheKey checks ImpersonatedUser first, so a SessionConfig carrying both resolves to basic:{user}. That's the ordering worth pinning.

Neo4j.Driver/Neo4j.Driver/Public/Auth/IAuthTokenManager.cs
```diff
@@ -22,6 +22,15 @@ namespace Neo4j.Driver;
 /// Common interface for components that can provide auth tokens. For pre-baked implementations of this interface,
```
AndyHeap-NeoTech, line 22: The IAuthTokenManager remarks won't be seen where users make the mistake

Public/Auth/IAuthTokenManager.cs:22-31. The content is good and correctly scoped. But an application author reaches this via ConfigBuilder.WithAuthTokenManager and AuthTokenManagers.*, and a <remarks> on the interface doesn't surface in IntelliSense at those call sites. Worth an <inheritdoc>-style pointer or a one-line cross-reference from WithAuthTokenManager.

Also worth considering, since SEC-03's residual exposure is now closed by documentation alone: a one-time Warn when a rotated token's principal differs from the previous one. It is silent misrouting that makes this dangerous, and a log line does not violate the testkit contract as I read it (test_cache_considers_all_driver_auth_equal asserts no new ROUTE request, not the absence of a warning) — though that's worth confirming against the stub harness before committing to it. Genuinely optional; the docs may be enough for a 6.x security response.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
RichardIrons-neo4j: Confirmed, and this one is a regression I introduced rather than something inherited. Fixed in 46311e31.

You are right about the mechanism and right about the direction, which is the part that matters. `ReleaseAsync` drops the connection from `_inUseConnections`, then awaits `OnReleaseAsync`, which calls `ClearConnectionAsync` and puts a RESET on the wire, and only then adds it to `_idleConnections`. So the gap is not a few instructions, it is a server round trip, and for that whole window the connection is in neither collection. The acquire side has the same shape around the liveness probe.

What makes it worse than the bug it replaced is the sign of the error. `CanUseHomeDbCache` needs `ssrDisabledCount == 0`, so under-reporting the one SSR-disabled connection turns the cache on in exactly the mixed cluster it is meant to stay off in. The counters I removed over-reported, because they counted from first borrow until destroy and never decremented on release, so they could only ever hold the cache off for too long. I swapped a fail-safe error for a fail-unsafe one and called it a fix, which is the opposite of what the finding was about.

The fix follows your suggestion. A connection is tracked from the end of a successful `InitAsync` until `DestroyConnectionAsync`, which is the same lifetime `_poolSize` already had, and membership in that set decides whether a count is owed: `TryAdd` and `TryRemove` returning false means no adjustment. That is what stops the drift, because the original defect was not the counters themselves but the asymmetry between incrementing on borrow and decrementing on destroy. A connection that fails init is never added, so it can never be subtracted, and the `-1` regression test from the earlier round still passes.

Keeping counters rather than counting the set on each read is safe because `SsrEnabled` is only ever written by `HelloResponseHandler.UpdateSsrEnabled` during the HELLO exchange, so it is settled inside `InitAsync` and immutable for the rest of the connection's life. A count taken when the connection joins cannot go stale.

That also answers your second point, since the reads are two `int` loads again instead of a locked snapshot of every bucket per acquisition.

There is a new test for the window itself: a validator that blocks inside `OnReleaseAsync` lets the test observe the pool mid-release. It reads 0 before the fix and 1 after.

One piece of your comment I have not done. Reading the enabled and disabled counts as a single consistent snapshot still is not guaranteed, because `ConnectionCausesCacheDisable` reads the two properties separately. The tear window is now two adjacent `int` loads rather than two full table snapshots, which is what the code did before this PR, so it is not a regression, but it is not airtight either. Closing it properly means a combined accessor on `IConnectionPool` and changing both consumers, and I read your requirement as conditional on keeping the enumeration approach. Happy to add it here if you would rather it were exact.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
RichardIrons-neo4j: Deleted, along with the `IConnectionPool` member, in 29df87e9.

Routing the cluster check through it was the other option and it does not work: the question is cluster-wide and a single pool cannot see the other pools, so `ClusterConnectionPool.ConnectionCausesCacheDisable` has to stay where it is. That leaves the per-pool version with nothing to do.

You are right that citing it as corroborating evidence and then leaving it in place was the wrong outcome. The PR body now explains that it was deleted rather than promoted.

Neo4j.Driver/Neo4j.Driver/Public/TrustManager.cs
RichardIrons-neo4j: Fair, and it was not a conscious decision before you raised it. Addressed in 61649eca.

I have kept the helper `internal` and documented the precondition on `ValidateServerCertificate` instead, so an external implementer is told that the missing certificate case is theirs to handle and that callers do not screen it out first. Widening to `protected` would have made the guard reusable, but it adds public surface for a five line check, and the thing an implementer actually needs is to know the obligation exists.

Renamed to `TryRejectMissingCertificate`. Your reading is right: it logs, so it should not look like a pure predicate.

Neo4j.Driver/Neo4j.Driver/Public/TrustManager.cs
RichardIrons-neo4j: Fixed in 61649eca. The summary now says that any certificate the peer presents is accepted without validation, and that a peer presenting none is still refused.

You have put your finger on the inconsistency in my own argument. I spent a paragraph of the PR body reasoning about whether narrowing this was breaking, and then left the documented contract saying the old thing.

Neo4j.Driver/Neo4j.Driver.Tests/Connector/Trust/TrustManagerTests.cs
RichardIrons-neo4j: All three added in a59ef60a: a certificate presented alongside `RemoteCertificateNotAvailable`, a null certificate with no policy errors, and a happy path.

The last one was the real gap. Your observation that a `ValidateServerCertificate` returning `false` unconditionally would have passed the whole file was accurate, so I checked the new tests against exactly that mutation: with the guard forced to refuse everything, the happy path fails and the rest still pass. Before the change nothing in the file failed.

The happy path uses `CreateInsecure` with a self signed certificate, since that is the only manager that can trust one. Chain, peer and cert trust all legitimately reject a self signed certificate, so a happy path for those would need a real trust anchor.

Neo4j.Driver/Neo4j.Driver.Tests/Routing/RoutingTableManagerTests.cs
RichardIrons-neo4j: Correct on both counts, fixed in dfd01252.

The seeded table was fresh, so the early return fired and the parameter was never read. I have kept that test, since as you say it does guard against the early return change coming back, and dropped the pool manager and address provider setups that were never reached.

The new test covers the stale case: a stale table plus `isCachedHomeDb: true`, asserting discovery is called with `""` and never with `db_alice`. I confirmed it fails if `isCachedHomeDb ? "" : database` is reduced to `database`, so the parameter's behaviour is now actually pinned rather than incidentally satisfied.

Neo4j.Driver/Neo4j.Driver.Tests/HomeDbCaching/HomeDbCacheKeyProviderTests.cs
RichardIrons-neo4j: You are right, and the reason is worse than tautological: `driverLevelAuthToken` appears only in the signature and is never read in the body, so nothing about that assertion could fail for any reason connected to precedence. Replaced in 15f42f21.

The replacement is the ordering you pointed at. A `SessionConfig` carrying both an impersonated user and its own token resolves to `basic:{user}`, and the test fails if the two branches are swapped in the provider. `ShouldTreatAllDriverLevelAuthTokensAsEqual` stays as the guard against sub-fix (a) returning.

Neo4j.Driver/Neo4j.Driver/Public/Auth/IAuthTokenManager.cs
RichardIrons-neo4j: Agreed on the substance, addressed in a444a34c. One correction on the specifics: there is no `ConfigBuilder.WithAuthTokenManager`. A manager reaches the driver through the `AuthTokenManagers` factories and the `GraphDatabase.Driver` overloads that take one, so those are the three parameter docs and the class summary I have cross referenced.

On the warning when a rotated token's principal changes, I would rather not put it in this PR. I agree with your reading that the testkit assertion is about ROUTE requests and would not be violated by a log line, and silent misrouting is the dangerous part. But it means comparing successive tokens on the acquisition path, and it is a new log surface that wants its own stub verification, which is a different shape of change from the rest of this response. I have written it up as a follow-up rather than dropping it.

AndyHeap-NeoTech (commented):
Looks like you missed the comment in the CustomAuthToken.cs file.
Also added a new finding, comment is in connectionPool.cs but the actual change is in a helper function.

Neo4j.Driver/Neo4j.Driver/Internal/ConnectionPool.cs
```diff
@@ -399,23 +379,50 @@ private async Task DestroyConnectionAsync(IPooledConnection conn)
...
+        }
+        finally
+        {
+            _poolMetricsListener?.ConnectionClosed();
+        }
+    }
+
+    private void StartTrackingConnection(IPooledConnection conn)
+    {
+        if (!_liveConnections.TryAdd(conn))
```
AndyHeap-NeoTech, line 397: This guard is never taken. [ConcurrentHashSet.TryAdd]
Neo4j.Driver/Neo4j.Driver/Internal/Util/ConcurrentHashSet.cs#L41-L44) returns `_items.GetOrAdd(item, _ => true)`, which is the value, not whether an insert happened - so it is always true, contradicting its own doc comment.

Fix is in the helper, one line: return `_items.TryAdd(item, true);`. The only other caller (ConnectionPool.cs:545) discards the return, so it's safe.

RichardIrons-neo4j: Confirmed, fixed in b881fefe. `GetOrAdd` hands back the value stored against the key, which is hardcoded to `true` here, so `TryAdd` reported success for an item that was already present and contradicted its own doc comment. Changed to `_items.TryAdd(item, true)` as you suggested.

Your caller analysis matches what I found: the line in `StartTrackingConnection` is the only place in the codebase that reads the result. `ConnectionPool.cs:545` discards it, the twenty-odd test call sites discard it, and the two remaining `TryAdd` hits are two argument `ConcurrentDictionary` calls on a different type, so nothing else changes behaviour.

Worth correcting something I claimed in the SSR counting thread. I said membership decides whether a count is owed because `TryAdd` and `TryRemove` returning false means no adjustment. Only half of that was true. `TryRemove` does report accurately, which is why the failed init case still could not produce a `-1` and why that regression test was passing for the right reason. The add side was never enforced, so the protection against counting a connection twice was decoration. It happens not to matter today, since `StartTrackingConnection` runs once per freshly created connection at the single creation site and nothing tracks the same object twice, but I wrote the guard as though it were load bearing and it was not.

There is now a `ConcurrentHashSetTests` alongside `ConcurrentOrderedSetTests`, covering both methods, so the contract is pinned at the level where it belongs rather than implicitly through the pool.

The defect predates this PR: it arrived in e540a444 in February 2023. Mine is just the first code to depend on the return value.

AndyHeap-NeoTech (approved):

PR #960, opened 2026-08-28
Don't clear testkit backend object store

RichardIrons-neo4j:
Follow-up to #957. `ObjectStore` is a process singleton but was cleared per connection, and testkit doesn't wait for the backend's teardown — so a closing connection could delete the next test's objects (in CI, driver `18428` vanished between its own `NewDriver` response and the following `NewSession`). Fixed by never clearing the store, which makes the first commit's semaphore unnecessary — the second commit reverts it.

### DIFF
--- a/Neo4j.Driver/Neo4j.Driver.TestKitBackend/Connection/TestkitConnectionHandler.cs
+++ b/Neo4j.Driver/Neo4j.Driver.TestKitBackend/Connection/TestkitConnectionHandler.cs
@@ -29,6 +29,7 @@ internal class TestkitConnectionHandler : ConnectionHandler
     private readonly IConnectionIdProvider _connectionIdProvider;
     private readonly ILoggingContextAccessor _loggingContextAccessor;
     private readonly ILogger _logger;
+    private readonly SemaphoreSlim _connectionTurn = new(1, 1);
 
     public TestkitConnectionHandler(
         ILifetimeScope rootScope,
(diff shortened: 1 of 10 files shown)


Neo4j.Driver/Neo4j.Driver.TestKitBackend/Connection/TestkitConnectionHandler.cs
```diff
@@ -47,6 +48,19 @@ public TestkitConnectionHandler(
     }
 
     public override async Task OnConnectedAsync(ConnectionContext connection)
+    {
+        await _connectionTurn.WaitAsync();
```
AndyHeap-NeoTech: I'm a bit concerned that this WaitAsync has no timeout, and that there is no logging. I _think_ that we could have the situation where a hung stub server would cause a cascade of test failures rather than just one. And that would happen without any logging.

Perhaps something like this would work?

`if (!await _connectionTurn.WaitAsync(TimeSpan.FromSeconds(30)))
{
    _logger.LogWarning("Previous connection has not finished tearing down; proceeding anyway");
}`

AndyHeap-NeoTech (approved):
The fix looks good. One comment about a potential behaviour change under failure conditions. 

And observation: there might be a tension with the singleton nature of IObjectStore and it being shared around. This kind of pattern often leads to small windows where state gets all messed up and pollutes between the non-singleton users.

AndyHeap-NeoTech (approved):
All good.

